PR #8, opened 2023-09-27
added parameters for not sending to NERSC cleanup

dylanmcreynolds:


orchestration/flows/bl832/move.py
```diff
@@ -105,25 +105,45 @@ def ingest_scicat(config: Config832, relative_path):
...
+        - Schedule a job to delete from spot832 in the future
+        - Schedule a job to delete from data832 in the future
+
+    The is_export_control and send_to_nersc flags are functionally identical, but
+    they are separate options at the beamlines, so we leave them as separate parameters
+    in case the desired behavior changes in the future.
+
+    :param file_path: path to file on spot832
+    :param is_export_control: if True, do not send to NERSC or ingest into SciCat
+    :param send_to_nersc: if True, send to NERSC or ingest into SciCat
```
Wiebke, line 123: Would change this to "send to NERSC and ingest into SciCat"

orchestration/flows/bl832/move.py
```diff
@@ -105,25 +105,45 @@ def ingest_scicat(config: Config832, relative_path):
...
     transfer_spot_to_data(relative_path, config.tc, config.spot832, config.data832)
 
     logger.info(f"Transferring {file_path} to spot to data")
-    transfer_data_to_nersc(relative_path, config.tc, config.data832, config.nersc832)
-    logger.info(
-        f"File successfully transferred from data832 to NERSC {file_path}. Task {task}"
-    )
 
-    ingest_scicat(config, relative_path)
+    if not is_export_control and send_to_nersc:
```
Wiebke, line 139: Is the user alerted somewhere that their data is not being transferred to NERSC and ingested, e.g. on experiment start or through visible ticks in a GUI?

This probably does not make sense on a per-file basis, but may be useful on a per-experiment-run basis.

requirements.txt
```diff
@@ -2,5 +2,6 @@ globus-sdk>=3.0
 httpx>=0.22.0
 python-dotenv
 prefect==2.7.9
+pydantic<2.0
```
Wiebke, line 5: Is this just to lock in dependencies from other packages?

dylanmcreynolds: Oh, I should explained. This old version of prefect forgot to pin pydnatic. So if you build a new enviornment today, you get 2.x version of pydantic, which breaks prefect.

dylanmcreynolds: Also, I looked briefly into upgrading prefect because of this, but they refactored the API enough that I can't do that now.

orchestration/flows/bl832/move.py
dylanmcreynolds: Kind of not our problem, but the user controls both settings right now. In the future, one of the settings will come automatically from the ALSHub

Wiebke: I see, makes sense.

PR #19, opened 2024-04-17
bump agent prefect version

### LINKED ISSUE #18: Underlying agent prefect version upgrade
We have started deploying prefect agent builds with prefect 2.14, but this agent code uses a much oder version of prefect. https://github.com/als-computing/splash_flows_globus/blob/main/Dockerfile

We should do the following:

- [ ] The dockerfile should build from prefect 2.14
- [ ] We should re-release this with a new tag so that we produce a docker image with a label.

dylanmcreynolds:
Fixes #18


dylanmcreynolds:
Code reviewed in person

PR #20, opened 2024-04-17
remove nersc client from config

dylanmcreynolds:


dylanmcreynolds:
reviewed in person with Raja

PR #22, opened 2024-05-02
change nersc7012 NERSC root_path

dylanmcreynolds:
This PR changes the root path configured for beamline 7012 at NERSC.


davidalexandershapiro (approved):
Great work

PR #26, opened 2024-09-13
ALCF update

davramov:
1. Updated `orchestration/globus/transfer.py` to force a string data type to be passed into the transfer request.
2. Updated `scripts/check_globus_compute.py` and `scripts/check_globus_transfer.py` to use Typer instead of Argparse for running with terminal commands.


dylanmcreynolds:
This pull request should not have so many commits in it. I think the branch you're working on, `alcf-update`, was created from a branch that had not been updated from the upstream main. We need to fix this before merging.

PR #28, opened 2024-09-20
ALCF Update

davramov:
Changed argparse to typer in scripts/check_globus_compute.py and check_globus_transfer.py. Also patched orchestration/globus/transfer.py to coerce transfer data into strings.

### DIFF
--- a/orchestration/globus/transfer.py
+++ b/orchestration/globus/transfer.py
@@ -94,9 +94,9 @@ def init_transfer_client(app: GlobusApp) -> TransferClient:
 def start_transfer(
     transfer_client: TransferClient,
     source_endpoint: GlobusEndpoint,
-    source_path: Path,
+    source_path: str,
     dest_endpoint: GlobusEndpoint,
-    dest_path: Path,
+    dest_path: str,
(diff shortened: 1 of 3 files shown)


orchestration/globus/transfer.py
```diff
@@ -115,7 +115,7 @@ def start_transfer(
             relative_path = item.relative_to(source_path.parent)
             tdata.add_item(str(item), os.path.join(dest_path, str(relative_path)))
     else:
-        tdata.add_item(source_path, dest_path)
+        tdata.add_item(str(source_path), str(dest_path))
```
dylanmcreynolds, line 118: If the issue is that the variables are `pathlin.Path` instances, then can we also change the type hints for the variables in the function call?

dylanmcreynolds (changes requested):

orchestration/globus/transfer.py
dylanmcreynolds: oh wait...

the typing is correct  on the function definition, it's that we reinitizlize source_path.

dest_path probably comes in as a str and doesn't need to be coerced

Issue #27, opened 2024-09-20
Mock Globus client should not inherit from real globus client


dylanmcreynolds:
https://github.com/als-computing/splash_flows_globus/blob/a2b20e1375ae11e58fbc3552a8745d85482bbc7c/orchestration/_tests/test_globus_flow.py#L104

dylanmcreynolds:
@davramov  can you address this?

davramov:
I have addressed in my PR: https://github.com/als-computing/splash_flows_globus/pull/29

PR #29, opened 2024-10-02
ALCF fixes: tiff_to_zarr.py dynamic pixelsize, init_tiff_to_zarr_globus_flow.py, fix for issue #27

davramov:
- Updated `tiff_to_zarr.py` to read pixelsize from the raw h5 file and convert to micrometer for the multiscales step.
- Updated `init_tiff_to_zarr_globus_flow.py` to account for the new change (passing in the h5 path to the flow) as well as automatically updating the Prefect Secret Block on the server with the latest `flow_id` and `func_id` once the script runs. Side note: I removed the time import from the function definition because I was getting a cryptic "`time.run()` is not defined error" despite only calling `time.time()`.
- Updated instructions for running `check_globus_compute.py` in the docstring.

---
  - https://app.asana.com/0/0/1208451498137898

### DIFF
--- a/scripts/init_tiff_to_zarr_globus_flow.py
+++ b/scripts/init_tiff_to_zarr_globus_flow.py
@@ -1,5 +1,5 @@
 from globus_compute_sdk import Client, Executor
-from prefect.blocks.system import Secret
+from prefect.blocks.system import Secret, JSON
 from prefect import task, flow, get_run_logger
 
 from check_globus_compute import check_globus_compute_status
@@ -18,7 +18,8 @@ def get_polaris_endpoint_id() -> str:
     return compute_endpoint_id
 
 
-def conversion_wrapper(rundir, h5_file_name, folder_path):
+# def conversion_wrapper(rundir, h5_file_name, folder_path):
+def conversion_wrapper(rundir, recon_path, raw_path):
     """
     Python function that wraps around the application call for Tomopy reconstruction on ALCF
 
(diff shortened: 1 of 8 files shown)


scripts/polaris/tiff_to_zarr.py
```diff
@@ -58,8 +71,8 @@ def main():
     backend = detect_cli_io_backend(file_paths)
     image = cli_input_to_ngff_image(backend, file_paths)
     # The scale and axis units are the same as the one printed in the reconstruction script
-    image.scale = {'z': 0.65, 'y': 0.65, 'x': 0.65}
-    image.axes_units = {'z': 'micrometer', 'y': 'micrometer', 'x': 'micrometer'}
+    image.scale = read_pixelsize_from_hdf5(args.raw_directory)
+    image.axes_units = {'x': 'micrometer', 'y': 'micrometer', 'z': 'micrometer'}
```
dylanmcreynolds, line 75: I think you can get units from the h5 file as well.

dylanmcreynolds (changes requested):

scripts/init_tiff_to_zarr_globus_flow.py
```diff
@@ -31,9 +31,6 @@ def conversion_wrapper(rundir, h5_file_name, folder_path):
     """
     import os
     import subprocess
-    import time
```
dylanmcreynolds, line 34: Was the issue that references time.run() at the import or the use of `time.time()`

That's super duper odd. I would be interested to see `print(time.time)` if it's no occuring during the import.

davramov: I couldn't recreate the issue locally but only when the Globus Flow called the function. There error listed specifically the line `start = time.time()`. What's odd to me is A) it was working before and B) I am calling time.time() and nowhere do I call time.run().

scripts/polaris/tiff_to_zarr.py
davramov: The pixel_size attribute seems to only store the numerical value but not the units. The name of the camera model is also listed in the h5 file (PCO_Edge), but not the units. The units are defined in a comment in one of Dula's Jupyter notebooks (`notebooks/microct/backend/ALS_recon_functions.py`)

Here are the attributes stored in `/measurement/instrument/detector/` in the h5 file:
```{'binning_x': array([1.]),
'binning_y': array([1.]),
'dark_field_value': array([100.]),
'delay_time': array([0.]),
'dimension_x': array([2560.]),
'dimension_y': array([2160.]),
'exposure_time': array([0.15, 0.15, 0.15, ..., 0.15, 0.15, 0.15]),
'model': b'PCO_Edge',
'pixel_size': array([0.00065]),
'temperature': array([25.])}```

dylanmcreynolds:
Fixes #27

dylanmcreynolds (approved):

PR #32, opened 2024-10-10
Uber 832 Flow Decider

davramov:
Added `orchestration/flows/bl832/decision_flow.py` that sets up the JSON block for storing which flows to run, and a Prefect flow that reads from that JSON block, and runs each flow set to True. I still need to define inputs for each flow.

### DIFF
--- a/orchestration/flows/bl832/decision_flow.py
+++ b/orchestration/flows/bl832/decision_flow.py
@@ -1,0 +1,58 @@
+from prefect import flow, task, get_run_logger
+from prefect.blocks.system import JSON
+from prefect.deployments.deployments import run_deployment
+
+
+@task(name="setup_decision_settings")
+def setup_decision_settings(alcf_recon: bool,
+                            nersc_recon: bool,
+                            nersc_move: bool) -> dict:
+    """
+    This task is used to define the settings for the decision making process of the BL832 beamline.
+    """
+    logger = get_run_logger()
+    logger.info(f"Setting up decision settings: alcf_recon={alcf_recon}, "
+                f"nersc_recon={nersc_recon}, nersc_move={nersc_move}")
+    settings = {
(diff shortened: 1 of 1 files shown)


orchestration/flows/bl832/decision_flow.py
```diff
@@ -0,0 +1,58 @@
...
+
+@flow(name="decision_flow")
+def decision_flow():
+    """
+    This flow is used to define the decision making process of the BL832 beamline.
+    """
+    logger = get_run_logger()
+    logger.info("Starting decision flow")
+    decision_settings = JSON.load("decision-settings")
+    for flow_name, run_flow in decision_settings.value.items():
```
dylanmcreynolds, line 44: Please us logger or delete print statements.

Another point here, this runs the flows sequenitally. We  want them run concurrently.

davramov: This push addresses this issue: https://github.com/als-computing/splash_flows_globus/pull/32/commits/edecb863a8ab1dfcb27048b1f55ac3e47a9b9891

orchestration/flows/bl832/decision_flow.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    This flow reads the decision settings and launches tasks accordingly.
+    """
+    logger = get_run_logger()
+    logger.info("Starting decision flow")
+    decision_settings = await JSON.load("decision-settings")
+
+    tasks = []
+    for flow_name, run_flow in decision_settings.value.items():
+        logger.info(f"{flow_name}")
```
dylanmcreynolds, line 57: One cool trick in python:

`logger.info(f"{flow_name=}")` automatically prints the variable name along with the value.

dylanmcreynolds (approved):

PR #33, opened 2024-10-16
832 Uber Flow Progress

davramov:
Updated `decision_flow.py` to:
- dynamically call all enabled flows
- populate each function call with the correct parameters using a structured Class
- data validation using pydantic
- error logging

Added a new pytest case to `orchestration/_tests/test_globus_flow.py` to test the uber flow.

Added a shell script to deploy the uber flow

### DIFF
--- a/orchestration/_tests/test_globus_flow.py
+++ b/orchestration/_tests/test_globus_flow.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import Any, Dict, List, Optional
 from uuid import UUID, uuid4
 import warnings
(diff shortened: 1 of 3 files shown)


orchestration/_tests/test_globus_flow.py
```diff
@@ -183,6 +191,54 @@ def get_run(self, run_id: UUID) -> Dict[str, Any]:
...
+    async def mock_run_deployment(*args, **kwargs):
+        return None
+
+    mocker.patch('prefect.deployments.deployments.run_deployment', new=mock_run_deployment)
+
+    # Mock asyncio.gather to avoid actual async task execution
+    async def mock_gather(*args, **kwargs):
+        return [None]
+
+    mocker.patch('asyncio.gather', new=mock_gather)
```
GBirkel, line 225: Would be nice to have a comment here explaining why we need to stub `asyncio.gather`?

GBirkel (approved):
After an interesting discussion with David, I think I understand this code a bit better!  (Still some studying to do with Prefect/Globus on my part.)

Marking as "approved" so it can move along...

PR #44, opened 2024-12-04
NERSC SFAPI Reconstruction Flow

davramov:
Setting this up as a draft PR for now while I iron out implementation details to match the ALCF workflow.

These two steps are currently working via sfapi and the container Raja built (registry.nersc.gov/als/tomorecon_nersc_mpi_hdf5):
- Reconstruction
- Tiff to Zarr

There are still at least a few things that need to be addressed before it is production ready:
- Pass into reconstruction.py the same file_path that `new_832_file_flow` and `alcf_recon_flow` expect (still using a legacy version of reconstruction.py that expects an input.txt file)
- File transfers back to data832
- Schedule Pruning
- Pytest cases


orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+        ABC: Abstract Base Class
+    """
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def reconstruct(
+        self,
+        file_path: str = "",
+        is_export_control: bool = False,
```
dylanmcreynolds, line 30: I think it's unlikely that these classes will need to know about export control. Whatever calls them will do a lot of things like not copying files before deciding not to reconstruct them.

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+        :param config: Optional configuration dictionary.
+        :return: True if successful, False otherwise.
+        """
+        pass
+
+    @abstractmethod
+    def build_multi_resolution(
+        self,
+        file_path: str = "",
+        is_export_control: bool = False,
```
dylanmcreynolds, line 46: same comment as above

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+    """
+
+    def __init__(self):
+        pass
+
+    def reconstruct(
+        self,
+        file_path: str = "",
+        is_export_control: bool = False,
+        config: Optional[Dict[str, Any]] = None
```
dylanmcreynolds, line 75: Is the Config not a class with it's own type definition?

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+    ) -> bool:
+        """
+        Use NERSC for tomography reconstruction
+        """
+        logger.info("Starting NERSC reconstruction process.")
+
+        if is_export_control:
```
dylanmcreynolds, line 112: Shouldn't the config be passed into the init of these classes?

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+    ) -> bool:
+        """Use NERSC to make multiresolution version of tomography results."""
+
+        if is_export_control:
+            logger.warning("File is export controlled; skipping NERSC multi-resolution task.")
+            return False
+        else:
+            logger.info("File is not export controlled; proceeding with NERSC multi-resolution task.")
+
+        if not config:
```
dylanmcreynolds, line 143: Shouldn't the config be passed into the init of these classes?

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+
+        nersc_multi_resolution_success = self._submit_nersc_multi_resolution_job(
+            file_path=file_path
+        )
+
+        logger.info(f"Was NERSC multi-resolution conversion successful: {nersc_multi_resolution_success}")
+
+        return nersc_multi_resolution_success
+
+    def _create_nersc_client(self) -> NerscClient:
```
dylanmcreynolds, line 154: This shouldn't be an instance method (self is never used.) I'd rather make this a static method or a separate function so that we can inject the client into this class.

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+            logger.error("NERSC credential files are missing.")
+            raise FileNotFoundError("NERSC credential files are missing.")
+
+        try:
+            return NerscClient(client_id_path, sfapi_key_path)
+        except Exception as e:
+            logger.error(f"Failed to create NERSC client: {e}")
+            raise e
+
+    def _submit_nersc_reconstruction_job(
```
dylanmcreynolds, line 173: This probably doesn't need to be a separte function from `reconstruct()`

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+        self,
+        file_path: str = None
+    ) -> bool:
+        """Submit a tomography reconstruction job to NERSC"""
+
+        if self.client is None:
+            logger.error("NERSC client is required for job submission.")
+            raise ValueError("NERSC client is required for job submission.")
+
+        load_dotenv()
```
dylanmcreynolds, line 183: `load_dotenv` should only be called once and can be done at the modeule level.

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+            logger.info("Submitting reconstruction job script to Perlmutter.")
+            job = self.client.perlmutter.submit_job(job_script)
+            job.complete()  # waits for job completion
+            logger.info("Reconstruction job completed successfully.")
+            return True
+        except Exception as e:
+            logger.error(f"Failed to submit or complete reconstruction job: {e}")
+            return False
+
+    def _submit_nersc_multi_resolution_job(
```
dylanmcreynolds, line 239: Probably doesn't need to be its own function.

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+        file_path: str = None,
+    ) -> bool:
+        """Submit a multi-resolution tomography job to NERSC"""
+
+        if self.client is None:
+            logger.error("NERSC client is required for job submission.")
+            raise ValueError("NERSC client is required for job submission.")
+
+        load_dotenv()
+        username = os.getenv("NERSC_USERNAME")
```
dylanmcreynolds, line 250: We really need to figure out how not to need user credentials for harbor.

orchestration/flows/bl832/tomography_hpc.py
```diff
@@ -0,0 +1,343 @@
...
+        password = os.getenv("NERSC_PASSWORD")
+
+        user = self.client.user()
+
+        home_path = f"/global/homes/{user.name[0]}/{user.name}"
+        scratch_path = f"/pscratch/sd/{user.name[0]}/{user.name}"
+        logger.info(home_path)
+        logger.info(scratch_path)
+
+        image_name = "tomorecon_nersc_mpi_hdf5@sha256:cc098a2cfb6b1632ea872a202c66cb7566908da066fd8f8c123b92fa95c2a43c"
```
dylanmcreynolds, line 197: This should be a configuration, not hard-coded.

dylanmcreynolds (changes requested):
Nice work.

I added a number of review comments. This is a really nice start, and makes the nersc-specific code easy to see and follow. If my comments were a little picky, it's because the code was easy to follow.

tomocagrahpy_hpt.py should be split into sepearate modules, something like: tomography_hpc.py, nersc.py and alcf.py so that no more code than necessary is imported (e.g. when running an ALCF job, you don't need to import NerscClient.)


Also, what about `job_controller.py` rather than `tomography_hpc.py`? Since this is under flows/bl832, tomography is implied. And we might some day write the job controllers for local (non-HPC) jobs.

orchestration/flows/bl832/tomography_hpc.py
davramov: I noticed that our als project on Harbor is marked as 'private'. Perhaps we can set up another project for images we don't mind being 'public' that could get around needing credentials? I can open a ticket at NERSC to see what they suggest

image

dylanmcreynolds: I see what you're saying, but we need to solve this somewhat globably across project.

orchestration/flows/bl832/job_controller.py
```diff
@@ -0,0 +1,287 @@
...
+    hpc_type: str = None
+) -> TomographyHPCController:
+    """
+    Factory function to retrieve the appropriate HPC controller.
+
+    :param hpc_type: The type of HPC environment, either 'ALCF' or 'NERSC'.
+    :return: An instance of TomographyHPCController.
+    :raises ValueError: If an invalid HPC type is provided.
+    """
+    if hpc_type == "ALCF":
```
dylanmcreynolds, line 261: Consider using an Enum for these values.

orchestration/flows/bl832/job_controller.py
```diff
@@ -0,0 +1,287 @@
...
+    def build_multi_resolution(
+        self,
+        file_path: str = "",
+    ) -> bool:
+        # uses Globus Compute to build multi-resolution tomography
+
+        pass
+
+
+class NERSCTomographyHPCController(TomographyHPCController):
```
dylanmcreynolds, line 84: NERSC-specific code can now be moved into `/flows/bl832/nersc.py`

dylanmcreynolds (changes requested):
This

orchestration/flows/bl832/job_controller.py
davramov: I tested this more thoroughly after moving `NERSCTomographyHPCController` definition into `/flows/bl832/nersc.py`, and unfortunately this leads to a circular import issue.

The issue arises because `get_controller()`/`HPC(Enum)` relies on importing the specific controller implementations. Since the flow scripts contain the controller-specific implementation and require importing `get_controller()` to kick off the flows, the circular import ensues. I can see if there is a cleaner way to approach this, as I get why we would want to define those classes in their respective nersc.py and alcf.py files. Otherwise, I will fallback to keeping the classes defined in `job_controller.py` to avoid the error.

orchestration/flows/bl832/nersc.py
```diff
@@ -0,0 +1,207 @@
...
+        folder_name = path.parent.name
+        file_name = path.stem
+
+        # Can't use this long term in production. Need to find a better way to handle credentials.
+        # Want to run this as the alsdev user
+        username = os.getenv("NERSC_USERNAME")
+        password = os.getenv("NERSC_PASSWORD")
+
+        job_script = f"""#!/bin/bash
+            #SBATCH -q preempt
```
davramov, line 86: Several queue things I have figured out:

- `debug` has no minimum time requirement, but using a different queue such as `preempt` requires allocating 2 hours minimum for a job (which I think is a bit overkill in our case, since recon should only take 5-10 min).
- Based on testing debug vs preempt, it seems like the the debug jobs get picked up much faster (perhaps due to smaller time constraint, as they are both medium priority).
- There is also the `interactive` queue, which has a high priority, but only 4 nodes x 2 jobs can be active simultaneously.

orchestration/flows/bl832/nersc.py
```diff
@@ -0,0 +1,207 @@
...
+            date
+            """
+
+        try:
+            logger.info("Submitting reconstruction job script to Perlmutter.")
+            job = self.client.perlmutter.submit_job(job_script)
+            job.complete()  # waits for job completion
+            logger.info("Reconstruction job completed successfully.")
+            return True
+        except Exception as e:
```
davramov, line 118: This will raise an error if the job is not visible right away as an 'active job' on perlmutter, despite having a job ID:

`11:00:25.867 | ERROR   | orchestration.flows.bl832.job_controller - Failed to submit or complete reconstruction job: Job not found: 33817933`

This is problematic because the job _is_ running (or at least in the queue), and it is important not to run the multires step until recon is complete

orchestration/flows/bl832/job_controller.py
```diff
@@ -0,0 +1,332 @@
...
+    try:
+        hpc_enum = HPC(hpc_str)
+    except ValueError:
+        raise ValueError(f"'{hpc_type}' is not a valid HPC") from None
+
+    # Map HPC enum members to corresponding controller constructors
+    controller_map: dict[HPC, Callable[[], TomographyHPCController]] = {
+        HPC.ALCF: lambda: ALCFTomographyHPCController(),
+        HPC.NERSC: lambda: NERSCTomographyHPCController(NERSCTomographyHPCController.create_nersc_client()),
+    }
```
davramov, line 309: Replace with if/else statements. Delay the import of the specific controller implementations. Move NERSCTomographyHPCController to nersc.py

If that doesn't work, try putting get_controller in its own method

dylanmcreynolds:
Now that you have a nice clean `NERSCTomographyHPCController`, you have something you can easy write unit tests for. I'd recommend mocking `NerscClient` and test that the calls to it are as expected. Also, are we still sure that we need `NerscClient`?

davramov:
> Now that you have a nice clean `NERSCTomographyHPCController`, you have something you can easy write unit tests for. I'd recommend mocking `NerscClient` and test that the calls to it are as expected. Also, are we still sure that we need `NerscClient`?

Sounds good to me, I will work on adding unit tests. I suspect we can replace references to `NerscClient` with the official `sfapi_client` `Client` class, and capture that logic in `create_nersc_client()` in `NERSCTomographyHPCController`. That could be a bit cleaner.

orchestration/flows/bl832/nersc.py
davramov: This issue was fixed by the sfapi_client team: https://github.com/NERSC/sfapi_client/issues/93

orchestration/flows/bl832/nersc.py
davramov: Bjoern let us know that we (under the 'als' project) have access to the `realtime` queue at NERSC, which should reduce latency. I had an issue using that queue with the following error:
`Error during job submission or completion: sbatch: error: Batch job submission failed: Invalid qos specification`

I opened a ticket and we are still waiting to hear from the help desk about the correct job parameters to use the queue:
https://nersc.servicenowservices.com/sp?sys_id=03d747c187a216101c6cea470cbb35e9&view=sp&id=ticket&table=incident

orchestration/flows/bl832/nersc.py
```diff
@@ -0,0 +1,297 @@
...
+    """
+    Implementation for a NERSC-based tomography HPC controller.
+
+    Submits reconstruction and multi-resolution jobs to NERSC via SFAPI.
+    """
+
+    def __init__(
+        self,
+        client: Client = None,
+        config: Optional[Config832] = None
```
dylanmcreynolds, line 34: Let's make this not optional

dylanmcreynolds (changes requested):
Let's do the following after discussion:
- add a big comment to the NerscClient saying that it's deprecated and will be removed when we refactor the ptychography code

- Remove initialization  of Config832 and sfapi Client from NERSCTomogrpahyHPCController, passing them down from flows to tasks as parameters

- Create unit tests for NERSCTomogrpahyHPCController methods

davramov:
- Added comments in `orchestration/nersc.py` indicating that NerscClient is deprecated and will be removed.
- Removed init of `Config832` and `sfapi_client` from `NERSCTomographyHPCController`, and instead inject them (no longer optional).
- Updated unit tests in `test_sfapi_flow.py`

davramov:
I met with Raja and we came up with a few next steps to wrap this up:

- Clean up home/scratch directories to make it production ready (in recon and multires implementations)
- Reconstructions need to be saved in the appropriate `scratch/` location on NERSC
- Transfer reconstructions back to `data832`
- Scheduling pruning
- Register Prefect flows/deployments for NERSC tomo reconstruction
- Create a new agent container on flow-prd to handle NERSC tomo reconstruction
- Come up with a game plan for sfapi credential reauthorization (can we request an SFAPI Client that will be valid for the whole cycle?)

orchestration/flows/bl832/prune.py
```diff
@@ -143,5 +143,33 @@ def prune_nersc832_alsdev_scratch(
...
+        relative_path=relative_path,
+        source_endpoint=source_endpoint,
+        check_endpoint=check_endpoint,
+        config=config)
+
+
+@flow(name="prune_nersc832_alsdev_pscratch_scratch")
+def prune_nersc832_alsdev_pscratch_scratch(
+        relative_path: str,
+        source_endpoint: GlobusEndpoint,
```
dylanmcreynolds, line 163: I just noticed this...doesn't this require that GlobusEntpoint be serializable? Does this prevent us from a call theis as a top-level flow?

orchestration/nersc.py
```diff
@@ -1,23 +1,30 @@
...
 # Temporary patch till the sfapi_client is updated
 from sfapi_client.jobs import JobSacct
-from sfapi_client.compute import Compute
+# from sfapi_client.compute import Compute
 JobSacct.model_rebuild()
 
 
 class NerscClient(Client):
+    '''
+    DEPRECATION WARNING: NerscClient is deprecated and will be removed when we refactor the ptychography code
```
dylanmcreynolds, line 26: Suggest we using the built-in warnings module for this:

```
import warnings

def old_function():
    warnings.warn(
        "old_function() is deprecated and will be removed in a future version. "
        "Use new_function() instead.",
        DeprecationWarning,
        stacklevel=2,  # Shows warning at caller level
    )
    # Function logic

```

orchestration/transfer_controller.py
```diff
@@ -0,0 +1,223 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+load_dotenv()
+
+
+Endpoint = TypeVar("Endpoint", bound=GlobusEndpoint)
+
+
+class BaseEndpoint(Protocol):
```
dylanmcreynolds, line 23: Protocols and ABCs do similar jobs in different ways. In part it's a matter of style which you choose. I prefer ABCs because I like inheritence and runtime enforcement. But I think we should be consistent in this PR and only use one or the other. Can you change this to ABCs?

FWIW, it doesn't make much sense to mark the methods in this class as `abstractmethod` for a Protocol.

orchestration/transfer_controller.py
```diff
@@ -0,0 +1,223 @@
...
+
+@dataclass
+class FileSystemEndpoint(BaseEndpoint):
+    root_path: str
+    name: str = "local"
+
+    def full_path(self, path_suffix: str) -> str:
+        if path_suffix.startswith("/"):
+            path_suffix = path_suffix[1:]
+        return f"{self.root_path.rstrip('/')}/{path_suffix}"
```
dylanmcreynolds, line 53: You're using root_path, but it's never set anywhere. It's defined as a property in the super class but there's no setter defined there.

orchestration/transfer_controller.py
```diff
@@ -0,0 +1,223 @@
...
+            logger.error(f"Failed to submit transfer: {e}")
+            return success
+        finally:
+            # Stop the timer and calculate the duration
+            elapsed_time = time.time() - start_time
+            logger.info(f"Transfer process took {elapsed_time:.2f} seconds.")
+            return success
+
+
+class SimpleTransfer(TransferController[FileSystemEndpoint]):
```
dylanmcreynolds, line 143: I am guessing taht this this would fail as rootpath is never set in FileSystemEndpoint. Can you either remove this (and save for a future PR) or right unit tests for it?

orchestration/transfer_controller.py
```diff
@@ -0,0 +1,223 @@
...
+    def copy(
+        self,
+        file_path: str = None,
+        source: Endpoint = None,
+        destination: Endpoint = None,
+    ) -> bool:
+        pass
+
+
+class GlobusTransfer(TransferController[GlobusEndpoint]):
```
dylanmcreynolds, line 79: This is really picky, but can we name this `GlobusTransferController`? Reviewing this I got a little mixed up between endpoints and transfer controllers.

orchestration/transfer_controller.py
dylanmcreynolds: This is really picky, but can we name this SimpleTransferController? Reviewing this I got a little mixed up between endpoints and transfer controllers.

dylanmcreynolds (changes requested):

dylanmcreynolds (approved):

orchestration/flows/bl832/prune.py
davramov: Interesting, you are right that GlobusEndpoint is not serializable but I haven't run into issues scheduling these prune calls. I added `to_dict` and `from_dict` methods to the GlobusEndpoint class so it can be serialized/deserialized if needed.

orchestration/transfer_controller.py
```diff
@@ -0,0 +1,293 @@
...
+        return self.name
+
+    def root_path(self) -> str:
+        """
+        Root path or base directory for this endpoint.
+        """
+        return self.root_path
+
+
+@dataclass
```
dylanmcreynolds, line 45: This shouldn't be a dataclass

dylanmcreynolds (changes requested):
One last tiny comment

Issue #47, opened 2024-12-18
3D Web App Next Steps


davramov:
After discussing with Elizabeth, Dula, and Dylan, we came up with a few next steps about visualizing reconstructions both at the beamline, and when users return home.

When the new `8.3.2 Viz Computer` arrives:
- [ ] Reinstall `Jupyter`/`ITK` visualization
- [ ] NFS mount the new viz computer's drive to `data832` and use that globus endpoint for transfers
- [ ] Host reconstructions via `Tiled` server running on the new BL computer
- [ ] Build `ITK` widget into a web app
    - [ ] Link from `SciCat` to the web app (no code) _and_ a `Jupyter` environment (deep dive)

Short term:
- [ ] Downsample raw data by a factor of 2 (bilinear) on `data832` before processing reconstruction (for more immediate results)
    - [ ] Run a Prefect worker on `data832` for downsampling
    - [ ] Register a `downsampling flow` on `flow-prd`
    - [ ] Call the `downsampling flow` _after_ `new_file_832` (once the raw data is on `data832`) and _before_ sending to an HPC facility (NERSC or ALCF)
    - [ ] Create queues for "downsampled" (high priority, fast throughput) and "full-scale" (slow, can sit in the queue) reconstructions
- [ ] Start building `ITK` widget into a web app

davramov:
Last week I tried getting the standalone web version of itkwidgets running locally, but couldn't, so I opened a git issue on their repository: https://github.com/InsightSoftwareConsortium/itkwidgets/issues/765

Paul responded, and their instructions worked for me. I was able to start an HTTP file server (with CORS enabled) serving an NGFF-ZARR directory, and then appended that path to the viewer URL as described:
`https://kitware.github.io/itk-vtk-viewer/app/?rotate=false&image=http://localhost:8000/rec20230606_152011_jong-seto_fungal-mycelia_flat-AQ_fungi2_fast.zarr`

Using this approach, if we host the zarr reconstructions on Tiled, we could dynamically generate the URL to visualize each dataset, and add the link in SciCat.

davramov:
[...]

PR #48, opened 2024-12-18
Issue 47: 3D Viz Next Steps

davramov:
I am starting this as a draft PR for the features outlined here: https://github.com/als-computing/splash_flows_globus/issues/47

This initial commit includes:
- a small 11MB scan for testing (`examples/tomo_scan.h5`)
- downsampling methods using Torch and OpenCV (`orchestration/flows/bl832/downsample.py`)
    - For small datasets (~11MB), Torch/OpenCV performs similarly (4 seconds)
    - For large datasets (~7GB), OpenCV works well, however, Torch appears to run into a memory issue (at least on my laptop) which prevents some of the images from downsampling correctly (`exchange/data`). To address this, I include a memory check, enabling Torch chunking. This slows down the downsampling (OpenCV: 80 seconds vs Torch: 123 seconds), but at least it works. I am keeping both versions here in case one works better in production.

### DIFF
--- a/orchestration/flows/bl832/view_recon_app/.env.example
+++ b/orchestration/flows/bl832/view_recon_app/.env.example
@@ -1,0 +1,1 @@
+DATA_PATH=/Users/david/Documents/data/tomo/scratch
--- a/orchestration/flows/bl832/view_recon_app/README.md
+++ b/orchestration/flows/bl832/view_recon_app/README.md
@@ -5,13 +5,69 @@
 This web application allows users to load their reconstructed volumes using Tiled Browser (in development) into the itk-vtk-viewer widget, all within the same web page.
 
 TODO:
-- Add a 'listener' mode that loads in the most recent tomographic reconstruction once processing is completed and the data is available.
-- Containerize the application to make it easily installable
-    - Are there deployments we want to support? Could we use this at other light sources?
 
+*   Add a 'listener' mode that loads in the most recent tomographic reconstruction once processing is completed and the data is available.
 
------------------------
-# Installation 
+---
+
+# Installation via Docker (Recommended)
+
+Before you begin, make sure you have Docker installed on your machine.
+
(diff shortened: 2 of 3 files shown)


davramov:
Yesterday I was able to get Seij's Bluesky TIled React component installed in my app via npm, and connected it to my "Select Dataset" button. There are still a few things to fix before it works properly:
- Test switching between datasets
- Update Tiled when a new dataset is available
- Fix the UI so there isn't a huge wrapper around the widget
- Load the bluesky icon correctly
- Close the Tiled Browser widget when the user presses "Select"

image

davramov:
While testing at the beamline, I noticed that I should make it easier to tell the app (Dockerfile) where Tiled should look for the data.

davramov:
A few steps left to make this perfect...

- currently the tiled server isn't picking up new zarrs in the hosted directory, and I have to restart the container each time a new scan is done. this is not ideal, it should update automatically
- set up the vis computer with a prefect agent + reconstruction job controller flow
    - Reconstruction on the new vis computer is very fast (2-3x faster than a full HPC node, 2-4 minutes vs 10 minutes)
    - tiff to zarr on the other hand is very slow (6x slower than HPC. 10 minutes vs nearly 60 minutes). I am not sure what exactly is causing this slowdown
    - best of both worlds: maybe we run automated reconstruction on the vis computer, then send it to an HPC facility for tiff to zarr, alternating between ALCF and NERSC to keep the throughput up

davramov:
A few things I've noticed after using the vis app some more:
- Tiled browser widget
    - broken image
    - the "Select" hover-over popup button is not wired into the select function. this could be confusing
- itk-vtk-viewer levels of detail
    - I haven't seen it actually load L0 or L1. Not sure if it's hitting some built-in browser limit, but I expect the vis PC with its specs should be able to handle it.

- A tutorial or info popup could be helpful

Issue #57, opened 2025-02-06
7.3.3 Flows


davramov:
Now that things have settled with infrastructure for 8.3.2, it's time to start extending this work to other beamlines.

Goals:
- Look into how file watching is currently implemented at 7.3.3 (meet with Wiebke, Dylan, and David)
- Create a new Prefect server
  - Maybe use Prefect v3 instead of v2 since we are starting fresh
- Once the Globus endpoint is running (should be soon), we can implement automated transfer code and other desired flows, similar to the 8.3.2
  - Move/store data long-term on NERSC
- Implement better authentication
  - OAuth2Proxy
  - KeyCloak -- Look into whether it makes more sense to have a different realm per beamline with user assignments or one realm with adjustable user permissions?
- See if there is any analysis code that BL scientists want to run after scans are completed that we can deploy as Prefect Flows, and call them automatically

PR #60, opened 2025-02-07
Issue 57: 7.3.3 Flows

davramov:
Creating this PR to start work on extending our Prefect workflows to beamline 7.3.3 once the Globus endpoint is set up as described here https://github.com/als-computing/splash_flows_globus/issues/57. I anticipate transferring data from 733 to NERSC as an initial step.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -1,5 +1,48 @@
 globus:
   globus_endpoints:
+
+    # 7.0.1.2 ENDPOINTS
+
+    nersc7012:
+      root_path: /global/cfs/cdirs/als/gsharing/data_mover/7012
+      uri: nersc.gov
+      uuid: df82346e-9a15-11ea-b3c4-0ae144191ee3
+      name: nersc7012
+
+    data7012:
+      root_path: /
+      uri: hpc.lbl.gov
+      uuid: 639c49be-604f-423c-9c5d-82a53afe1bf1
+      name: data7012
+
+    # 7.3.3 ENDPOINTS
+
+    nersc733_alsdev:
+      root_path: /global/cfs/cdirs/als/data_mover/8.3.2/
+      uri: nersc.gov
+      uuid: d40248e6-d874-4f7b-badd-2c06c16f1a58
+      name: nersc733_alsdev
+
+    nersc733_alsdev_raw:
+      root_path: /global/cfs/cdirs/als/data_mover/8.3.2/raw
+      uri: nersc.gov
+      uuid: d40248e6-d874-4f7b-badd-2c06c16f1a58
+      name: nersc733_alsdev_raw
+
+    data733:
+      root_path: /
+      uri: data733.lbl.gov
+      uuid: tbd
+      name: data733
+
+    data733_raw:
+      root_path: /data/raw
+      uri: data733.lbl.gov
+      uuid: tbd
+      name: data733_raw
+
+    # 8.3.2 ENDPOINTS
+
     spot832:
       root_path: /
       uri: spot832.lbl.gov
@@ -89,18 +132,6 @@ globus:
       uri: nersc.gov
       uuid: df82346e-9a15-11ea-b3c4-0ae144191ee3
       name: nersc832
-
-    nersc7012:
-      root_path: /global/cfs/cdirs/als/gsharing/data_mover/7012
-      uri: nersc.gov
-      uuid: df82346e-9a15-11ea-b3c4-0ae144191ee3
-      name: nersc7012
-
-    data7012:
-      root_path: /
-      uri: hpc.lbl.gov
-      uuid: 639c49be-604f-423c-9c5d-82a53afe1bf1
-      name: data7012
 
   globus_apps:
     als_transfer:
--- a/orchestration/_tests/test_globus_flow.py
+++ b/orchestration/_tests/test_globus_flow.py
@@ -81,10 +81,6 @@ class MockGlobusComputeClient:
 class MockSecret:
     value = str(uuid4())
 
-# ----------------------------
-# Tests for 733
-# ----------------------------
-
 
 # ----------------------------
 # Tests for 733
@@ -117,7 +113,7 @@ class MockConfig733:
 def test_process_new_733_file(mocker: MockFixture) -> None:
     """
     Test the process_new_733_file flow from orchestration.flows.bl733.move.
-    
+
     This test verifies that:
       - The get_transfer_controller function is called (patched) with the correct parameters.
       - The returned transfer controller's copy method is called with the expected file path,
--- a/orchestration/_tests/test_globus_flow.py
+++ b/orchestration/_tests/test_globus_flow.py
@@ -51,6 +51,121 @@ class MockEndpoint:
         self.uri = f"mock_endpoint_uri_{self.uuid}"
 
 
+# Mock the Client class to avoid real network calls
+class MockGlobusComputeClient:
+    def __init__(self, *args, **kwargs):
+        # No real initialization, as this is a mock
+        pass
+
+    def version_check(self):
+        # Mock version check to do nothing
+        pass
+
+    def run(self, *args, **kwargs):
+        # Return a mock task ID
+        return "mock_task_id"
+
+    def get_task(self, task_id):
+        # Return a mock task response
+        return {
+            "pending": False,
+            "status": "success",
+            "result": "mock_result"
+        }
+
+    def get_result(self, task_id):
+        # Return a mock result
+        return "mock_result"
+
+
+class MockSecret:
+    value = str(uuid4())
+
+# ----------------------------
+# Tests for 733
+# ----------------------------
+
+
+# ----------------------------
+# Tests for 733
+# ----------------------------
+
+class MockConfig733:
+    def __init__(self) -> None:
+        """
+        Dummy configuration for 733 flows.
+        """
+        # Create mock endpoints
+        self.endpoints = {
+            "data733_raw": MockEndpoint(root_path="mock_data733_raw_path", uuid_value=str(uuid4())),
+            "nersc733_alsdev_raw": MockEndpoint(root_path="mock_nersc733_alsdev_raw_path", uuid_value=str(uuid4())),
+        }
+
+        # Define mock apps
+        self.apps = {
+            "als_transfer": "mock_als_transfer_app"
+        }
+
+        # Use the mock transfer client instead of the real TransferClient
+        self.tc = MockTransferClient()
+
+        # Set attributes for easy access
+        self.data733_raw = self.endpoints["data733_raw"]
+        self.nersc733_alsdev_raw = self.endpoints["nersc733_alsdev_raw"]
+
+
+def test_process_new_733_file(mocker: MockFixture) -> None:
+    """
+    Test the process_new_733_file flow from orchestration.flows.bl733.move.
+    
+    This test verifies that:
+      - The get_transfer_controller function is called (patched) with the correct parameters.
+      - The returned transfer controller's copy method is called with the expected file path,
+        source, and destination endpoints from the provided configuration.
+
+    Parameters:
+        mocker (MockFixture): The pytest-mock fixture for patching and mocking objects.
+    """
+    # Import the flow to test.
+    from orchestration.flows.bl733.move import process_new_733_file
+
+    # Patch the Secret.load and init_transfer_client in the configuration context.
+    with mocker.patch('prefect.blocks.system.Secret.load', return_value=MockSecret()):
+        mocker.patch(
+            "orchestration.flows.bl733.config.transfer.init_transfer_client",
+            return_value=mocker.MagicMock()  # Return a dummy TransferClient
+        )
+        from orchestration.flows.bl733.config import Config733
+
+    # Instantiate the dummy configuration.
+    mock_config = Config733()
+
+    # Generate a test file path.
+    test_file_path = f"/tmp/test_file_{uuid4()}.txt"
+
+    # Create a mock transfer controller with a mocked 'copy' method.
+    mock_transfer_controller = mocker.MagicMock()
+    mock_transfer_controller.copy.return_value = True
+
+    # Patch get_transfer_controller where it is used in process_new_733_file.
+    mocker.patch(
+        "orchestration.flows.bl733.move.get_transfer_controller",
+        return_value=mock_transfer_controller
+    )
+
+    # Execute the flow with the test file path and dummy configuration.
+    result = process_new_733_file(file_path=test_file_path, config=mock_config)
+
+    # Verify that the transfer controller's copy method was called exactly once.
+    assert mock_transfer_controller.copy.call_count == 1, "Transfer controller copy method should be called exactly once"
+    assert result is None, "The flow should return None"
+
+
+# ----------------------------
+# Tests for 832
+# ----------------------------
+
+
 class MockConfig832():
     def __init__(self) -> None:
         # Mock configuration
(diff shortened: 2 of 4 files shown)


init_work_pools.sh
```diff
@@ -0,0 +1,125 @@
...
+    try:
+        subprocess.run(
+            ["prefect", "work-pool", "inspect", pool],
+            check=True,
+            capture_output=True,
+        )
+        print(f"[Init:{beamline}] Work pool '{pool}' already exists.")
+    except subprocess.CalledProcessError:
+        print(f"[Init:{beamline}] Creating work pool: {pool}")
+        subprocess.run(["prefect", "work-pool", "create", pool, "--type", "process"], check=True)
```
kkamdin, line 119: maybe this is standard but I think the python injection here and above makes this script a little hard to read and maintain. For example, no standard python syntax highlighting, and you have to be extra careful about line indents if you need to edit. 

Would we be okay moving these snippets into a stand alone scripts in the `scripts/ `folder and calling them here via `python3 scripts/setup_work_pools.py `or similar?

orchestration/_tests/test_globus_flow.py
```diff
@@ -51,6 +51,123 @@ def __init__(self, root_path, uuid_value=None):
         self.uri = f"mock_endpoint_uri_{self.uuid}"
 
 
+# Mock the Client class to avoid real network calls
```
kkamdin, line 54: why is the tests folder underscored?

also wow check out that dependency injection

orchestration/_tests/test_globus_flow.py
```diff
@@ -51,6 +51,123 @@ def __init__(self, root_path, uuid_value=None):
...
+            "result": "mock_result"
+        }
+
+    def get_result(self, task_id):
+        # Return a mock result
+        return "mock_result"
+
+
+class MockSecret:
+    value = str(uuid4())
```
kkamdin, line 82: I think it's best practice to have deterministic behavior for testing, but maybe there's a reason we're using a different one here each time. I don't think you can set the seed for uuid4 so maybe just pick a value and set it here?

if it needs to be real uuid4:
`value = "550e8400-e29b-41d4-a716-446655440000"`

if you need a string
`value = "my-mock-secret-acbd-1234"`

orchestration/_tests/test_globus_flow.py
```diff
@@ -51,6 +51,123 @@ def __init__(self, root_path, uuid_value=None):
...
+class MockConfig733:
+    def __init__(self) -> None:
+        """
+        Dummy configuration for 733 flows.
+        """
+        # Create mock endpoints
+        self.endpoints = {
+            "data733_raw": MockEndpoint(root_path="mock_data733_raw_path", uuid_value=str(uuid4())),
+            "nersc733_alsdev_raw": MockEndpoint(root_path="mock_nersc733_alsdev_raw_path", uuid_value=str(uuid4())),
+        }
```
kkamdin, line 98: oh now I see why you're generating uuids, it's because you want different ones for different endpoints.

You could do something like this:

```
class MockSecret:
    @staticmethod
    def for_endpoint(endpoint_name):
        # Create deterministic UUID based on endpoint name
        namespace = uuid.UUID('12345678-1234-5678-1234-123456789012')
        return str(uuid.uuid5(namespace, endpoint_name))
```

then usage:
```
"data733_raw": MockEndpoint(root_path="mock_data733_raw_path", uuid_value=MockSecret.for_endpoint("data733_raw")),
```
so they're different but between endpoints but still the same test-to-test

orchestration/_tests/test_globus_flow.py
```diff
@@ -51,6 +51,123 @@ def __init__(self, root_path, uuid_value=None):
...
+        source, and destination endpoints from the provided configuration.
+
+    Parameters:
+        mocker (MockFixture): The pytest-mock fixture for patching and mocking objects.
+    """
+    # Import the flow to test.
+    from orchestration.flows.bl733.move import process_new_733_file
+
+    # Patch the Secret.load and init_transfer_client in the configuration context.
+    with mocker.patch('prefect.blocks.system.Secret.load', return_value=MockSecret()):
```
kkamdin, line 129: this is cool to see an example of. I haven't used the pytest mocker before.

orchestration/_tests/test_globus_flow.py
```diff
@@ -51,6 +51,123 @@ def __init__(self, root_path, uuid_value=None):
...
+    mocker.patch(
+        "orchestration.flows.bl733.move.schedule_prefect_flow",
+        return_value=None
+    )
+
+    # Instantiate the dummy configuration.
+    mock_config = Config733()
+
+    # Generate a test file path.
+    test_file_path = f"/tmp/test_file_{uuid4()}.txt"
```
kkamdin, line 146: are these tests running in a local environment or in a docker container / as a github action? I think i'd recommend making this file path deterministic too but maybe I'm missing something.

orchestration/_tests/test_globus_flow.py
```diff
@@ -51,6 +51,123 @@ def __init__(self, root_path, uuid_value=None):
...
+    result = process_new_733_file(file_path=test_file_path, config=mock_config)
+
+    # Verify that the transfer controller's copy method was called exactly once.
+    assert mock_transfer_controller.copy.call_count == 1, "Transfer controller copy method should be called exactly once"
+    assert result is None, "The flow should return None"
+
+
+# ----------------------------
+# Tests for 832
+# ----------------------------
```
kkamdin, line 168: I think it would be nice to move the tests for each beamline into their own folder with this directory structure.

```
orchestration/
├── flows/
│   ├── bl733/
│   │   ├── __init__.py
│   │   ├── move.py
│   └── bl832/
│       ├── __init__.py
│   │   ├── move.py
└── tests/
    ├── test_bl733/
    │   ├── __init__.py
    │   ├── test_move.py
    └── test_bl832/
        ├── __init__.py
        └── test_move.py
```

That way we can also put your `MockGlobusComputeClient` in a shared location at the test level or a in a shared pytest fixture, and you can run tests for each beamline separately more easily. Would that work? What do you think?

I don't think it needs to be part of this PR though. Just a follow up after this is merged.

orchestration/flows/bl733/dispatcher.py
```diff
@@ -0,0 +1,56 @@
...
+    is_export_control: bool = False,
+    config: Optional[Union[dict, Any]] = None,
+) -> None:
+    """
+    Dispatcher flow for BL733 beamline that launches the new_733_file_flow.
+
+    :param file_path: Path to the file to be processed.
+    :param is_export_control: Flag indicating if export control measures should be applied.
+                              (Not used in the current BL733 processing)
+    :param config: Configuration settings for processing.
```
kkamdin, line 22: does this function also need to be updated after PR #62? (i saw the commend above in the config.py file)

orchestration/flows/bl733/dispatcher.py
```diff
@@ -0,0 +1,56 @@
+import logging
+from prefect import flow
+from typing import Optional, Union, Any
+
+from orchestration.flows.bl733.move import process_new_733_file
+
+logger = logging.getLogger(__name__)
```
kkamdin, line 7: do you need to use the prefect logger to get these logs in the prefect ui?

orchestration/flows/bl733/dispatcher.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@flow(name="dispatcher", flow_run_name="dispatcher-{file_path}")
+def dispatcher(
+    file_path: Optional[str] = None,
+    is_export_control: bool = False,
+    config: Optional[Union[dict, Any]] = None,
+) -> None:
+    """
+    Dispatcher flow for BL733 beamline that launches the new_733_file_flow.
```
kkamdin, line 17: is "Dispatcher flow" a prefect thing? For my education can you say more about why you use this function that to my naive eyes looks like a wrapper?

orchestration/flows/bl733/dispatcher.py
```diff
@@ -0,0 +1,56 @@
...
+            file_path=file_path,
+            config=config
+        )
+        logger.info("Dispatcher flow completed successfully.")
+    except Exception as e:
+        logger.error(f"Error during processing in dispatcher flow: {e}")
+        raise
+
+
+if __name__ == "__main__":
```
kkamdin, line 55: is this main block for piecemeal testing or is this really used when deployed as `python dispatcher.py`

orchestration/flows/bl733/move.py
```diff
@@ -0,0 +1,187 @@
...
+) -> bool:
+    """
+    Prune (delete) data from a globus endpoint.
+    If days_from_now is 0, executes pruning immediately.
+    Otherwise, schedules pruning for future execution using Prefect.
+    Args:
+        file_path (str): The path to the file or directory to prune
+        source_endpoint (GlobusEndpoint): The globus endpoint containing the data
+        check_endpoint (Optional[GlobusEndpoint]): If provided, verify data exists here before pruning
+        days_from_now (datetime.timedelta): Delay before pruning; if 0, prune immediately
```
kkamdin, line 34: possibly misleading variable name -- it's actually in seconds right?

kkamdin (changes requested):
I did a fairly thorough review but then got to some comments about how this PR depends on another PR down in move.py and the other flows/bl733 files so I didn't review those super thoroughly... I'll definitely take another pass but wanted to submit the comments so far.

Let me know if you prefer to pair to go over my questions/comments, I know typing in replies and going back and forth can be a big time sink.

create_deployments_733_2.19.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Create work pools. If a work pool already exists, it will throw a warning but that's no problem
+prefect work-pool create 'dispatcher_733_pool'
+prefect work-pool create 'new_file_733_pool'
+prefect work-pool create 'prune_733_pool'
+
+
+# dispatcher_733_pool
+    # in docker-compose.yaml:
+    # command: prefect agent start --pool "dispatcher_733_pool"
```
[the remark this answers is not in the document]
davramov: The docker-compose.yaml lives in the `als_ansible` repo:
https://github.com/als-computing/als_ansible/pull/4/files#diff-e91f492bf5b7084b0ccf76dda4b6f6983a29dd03f16e87bdaa535c013c5c16c3

I also want to move these `create_deployment` scripts to a new folder `scripts/legacy/`

orchestration/flows/bl733/dispatcher.py
davramov: This is a pattern we used for 8.3.2, where the "dispatcher" orchestrates the different data tasks. Right now, all it is set up to do is file movement, but down the line, we can add additional steps like data processing, that are triggered once file movement is complete

init_work_pools.sh
davramov: I agree, splitting this up between shell and python code would be ideal

orchestration/_tests/test_globus_flow.py
davramov: I believe Dylan chose the "underscore" style for tests, probably to indicate that this is for development, and not meant to be used in other parts of the codebase

orchestration/flows/bl733/dispatcher.py
davramov: Yes, I will want to change the typing to accept a Config733 class built off the `BeamlineConfig` ABC once it is merged
https://github.com/als-computing/splash_flows/pull/62/files#diff-ea82c0df0204124b98fbc16a7abbf08f17d0268ab2e439fca4ae2f8a1e2098dc

docs/mkdocs/docs/bl733.md
```diff
@@ -0,0 +1,218 @@
+# Beamline 7.3.3 Flows
+
+This page documents the workflows supported by Splash Flows Globus at [ALS Beamline 7.3.3 (SAXS/WAXS/GISAXS)](https://saxswaxs.lbl.gov/user-information). Beamline 7.3.3 supports hard x-ray scattering techniques include small- and wide-angle x-ray scattering (SAXS/WAXS) and grazing-incidence SAXS/WAXS (GISAXS/GIWAXS).
+
+## Diagrams
+
+### Sequence Diagram
```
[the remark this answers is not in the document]
davramov: I can look into tweaking this with Mermaid...

init_work_pools.sh
davramov: I can rewrite this in python, and just make the prefect deploy call a subprocess command

orchestration/flows/bl733/move.py
davramov: I see, so the docstring wasn't updated to reflect that days_from_now expects a float, not datetime.

init_work_pools.sh
```diff
@@ -0,0 +1,125 @@
...
+fi
+
+# If PREFECT_API_URL is already defined in the container’s environment, it will use that value.
+# If not, it falls back to the default http://prefect_server:4200/api.
+: "${PREFECT_API_URL:=http://prefect_server:4200/api}"
+
+echo "[Init:${BEAMLINE}] Using prefect.yaml at $PREFECT_YAML"
+echo "[Init:${BEAMLINE}] Waiting for Prefect server at $PREFECT_API_URL..."
+
+# Wait for Prefect server to be ready, querying the health endpoint
```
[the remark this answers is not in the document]
davramov: This is to make sure we don't try to create work pools or register flows before the Prefect server starts up, in cases where this script is set to run as startup service in the docker compose.

orchestration/_tests/test_globus_flow.py
davramov: I incorporated this with my latest push

orchestration/_tests/test_globus_flow.py
davramov: These are the pytests that are run as part of the github actions. Maybe as part of https://github.com/als-computing/splash_flows/issues/85 we can change the naming

orchestration/flows/bl733/dispatcher.py
davramov: Prefect has `get_run_logger` that's compatible with python logging, which will emit logs to the server ui. There's also a `log_prints` arg in the flow decorator that will send print statements to the ui as well.

https://docs.prefect.io/v3/how-to-guides/workflows/add-logging

docs/mkdocs/docs/bl733.md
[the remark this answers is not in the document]
davramov: For the sequence diagram, I made it have fewer horizontal nodes so it fits vertically. I tried coercing the font size to be larger, but it just wasn't playing nice.

config.yml
```diff
@@ -1,5 +1,48 @@
...
+
+    data7012:
+      root_path: /
+      uri: hpc.lbl.gov
+      uuid: 639c49be-604f-423c-9c5d-82a53afe1bf1
+      name: data7012
+
+    # 7.3.3 ENDPOINTS
+
+    nersc733_alsdev:
```
GBirkel, line 20: Is it too late in the development process to propose using the beamline and general location as a prefix instead of a suffix when defining the endpoints?  e.g.

bl733-nersc-alsdev,
bl733-nersc-alsdev_raw
bl733-als-data733
bl733-als-data733_raw

(Sorry I didn't propose this much sooner.  We can always put this in a ticket instead.)

orchestration/flows/bl733/move.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # TODO: Ingest file path in SciCat
+    # Waiting for PR #62 to be merged (scicat_controller)
+
+
+@flow(name="move_733_flight_check", flow_run_name="move_733_flight_check-{file_path}")
+def move_733_flight_check(
+    file_path: str = "test_directory/test.txt",
+):
+    """Please keep your arms and legs inside the vehicle at all times."""
```
GBirkel, line 190: I love this :D

config.yml
davramov: I like that! It would be easy to change it now for 733, and follow that convention for additional beamlines. Switching 832 would be more work, and maybe belongs in a ticket.

Issue #59, opened 2025-02-06
Flows for moving back forth between cfs and hpss


dylanmcreynolds:
Now that we have a working implementation of the sfapi in this repo, we can start looking at other things we can do with it at NERSC.

We are using the CFS for all strorage. But much of our data should be moved to tape. That system is called HPSS. Software at NERSC called `hsi` is the way of moving data between the two file systems.  

While NERSC has a globus DTN for HPSS, it doesn't work for our use case because we depend on the `alsdev` collab account as a type of service account for our jobs, and there is no concept of collab accounts in HPSS. All data put there would be granted permissions for our personal accounts.

So, now that we have `sfapi` from prefect, we can start using HPSS.

Some use cases:

New file created:
* Data is transferred from beamline to CFS (globus)
* Data is transferred from CFS to HFS (HSI via sfapi)
* Information about location on both CFS and HPSS is stored in scicat for the dataset
* Prune job is setup in the future (6 months? 1,2 years) to delete from CFS. (Globus?)

Prune job:
* Data is deleted from CFS
* Scicat is updated

Recovery job
* Data is moved from HPSS to CFS
* Scicat is updated

Assigning David to start thinking about.

davramov:
I'm taking the time to think about this before jumping into the code since there are a few big pieces.

**HPSS** 
Some considerations from the [HPSS documentation](https://docs.nersc.gov/filesystems/archive/):

- Users should aim for file sizes between 100 GB and 2 TB. Avoid small files and huge files.
- The command-line tools `hsi` and `htar` allow you to move files in and out and address the need for grouping. The commands `tar` and `split` can help break up large files. `htar` is used to put bundles of files into HPSS
- Execute `hsi` commands from any Perlmutter login node or a Data Transfer Node by either typing `hsi <command>` or `hsi` 

Considering the first point, maybe we should avoid moving individual scans to HPSS, and instead wait until after a beamtime/experiment is completed, and then bundle and send it using [`htar`](https://docs.nersc.gov/filesystems/htar/). I am not sure about a simple way to keep track of when to copy projects to tape. We could consider scheduling something to run every week.

Either way, I can add a new class `HPSSTransferController()` in `orchestration/transfer_controller.py` that implements the CFS -> HPSS -> CFS copy actions using `hsi` and `htar` in a Slurm script scheduled with SFAPI.

**SFAPI**
[...]

PR #62, opened 2025-02-11
Issue 59: Transfer Flows between CFS and HPSS

davramov:
I am opening this PR to address issue https://github.com/als-computing/splash_flows_globus/issues/59.

Key components in this initial commit:
- Includes a new transfer controller `CFSToHPSSTransferController()` with logic for handling single files vs directories using `SFAPI`, `Slurm`, `hsi`, and `htar` based on HPSS best practices.
- Moves `create_sfapi_client()` to a new file called `orchestration/sfapi.py`. Now it can be easily accessed by multiple components, including the NERSC tomography workflow
- Includes new documentation in MkDocs for HPSS.
- Added an HPSS endpoint to config.yml that we can use to define the root storage location.
- Updates `orchestration/_tests/test_sfapi_flow.py` to reflect the new location of `create_sfapi_client().`

 
This is still a WIP and requires thorough testing, and there are a few outstanding tasks:
- Add logic to the new `HPSSToCFSTransferController()` class (currently is just a placeholder)
- Incorporate the HPSS transfer into our Prefect Flows.
- We need a strategy for storing the `SFAPI` id/key pair. Our application must make it easy to update these values since they expire fairly quickly.
- Refactor all of the `TransferControllers` to accept a generic `config=Config` rather than `config=Config832`, which will help generalize our codebase as we extend support to additional beamlines.
- Build a new script `orchestration/pruning_controller.py` that centralizes and implements Pruning logic for:
    - HPSS as a minimum for this PR, and then start thinking about:
    - Globus Endpoints
    - Local File System Endpoints
- Store HPSS file paths in SciCat

### DIFF
--- a/docs/mkdocs/docs/hpss.md
+++ b/docs/mkdocs/docs/hpss.md
@@ -1,0 +1,179 @@
+# Working with The High Performance Storage System (HPSS) at NERSC
+
+HPSS is the tape-based data storage system we use for long term storage of experimental data at the ALS. Tape storage, while it may seem antiquated, is still a very economical and secure medium for infrequently accessed data as tape does not need to be powered except for reading and writing. This requires certain considerations when working with this system.
+
+In `orchestration/transfer_controller.py` we have included two transfer classes for moving data from CFS to HPSS and vice versa (HPSS to CFS). We are following the [HPSS best practices](https://docs.nersc.gov/filesystems/HPSS-best-practices/) outlined in the NERSC documentation.
+
+HPSS is intended for long-term storage of data that is not frequently accessed, and users should aim for file sizes between 100 GB and 2 TB. Since HPSS is a tape system, we need to ensure storage and retrieval commands are done efficiently, as it is a mechanical process to load in a tape and then scroll to the correct region on the tape.
+
+While there are Globus endpoints for HPSS, the NERSC documentation recommends against it as there are certain conditions (i.e. network disconnection) that are not as robust as their recommended HPSS tools `hsi` and `htar`, which they say is the fastest approach. Together, these tools allow us to work with the HPSS filesystem and carefully bundle our projects into `tar` archives that are built directly on HPSS.
+
+## Working with `hsi`
+
+We use `hsi` for handling individual files on HPSS. [Here is the official NERSC documentation for `hsi`.](https://docs.nersc.gov/filesystems/hsi/)
+
+
+**Login to HPSS using `hsi`**
+
+```
+nersc$ hsi
+```
+
+
+**Common `hsi` commands**
+```
+hsi ls: show the contents of your HPSS home directory
+hsi mkdir [new_dir]: create a remote directory in your home
+hsi put [local_file_name]: Transfer a single file into HPSS with the same name
+hsi put -R [local_directory]: Transfer a directory tree into HPSS, creating sub-dirs when needed
+hsi get [/path/to/hpss_file]: Transfer a single file from HPSS into the local directory without renaming
+hsi rm [/path/to/hpss_file]: Prune a file from HPSS
+hsi rm -r [/path/to/hpss_file]: Prune a directory from HPSS
+hsi rmdir /path/to/my_hpss_dir/: Prune an empty directory
+
+```
+
+**Examples**
+
+Find files that are more than 20 days old and redirects the output to the file temp.txt:
+
+```
+hsi -q "find . -ctime 20" > temp.txt 2>&1
+```
+
+## Working with `htar`
+
+We can use `htar` to efficiently work with groups of files on HPSS. The basic syntax of `htar` is similar to the standard `tar` utility:
+
+```
+htar -{c|K|t|x|X} -f tarfile [directories] [files]
+
+-c : Create
+-K : Verify existing tarfile in HPSS
+-t : List
+-x : Extract
+-X : re-create the index file for an existing archive
+```
+
+You cannot add or append files to an existing htar file. The following examples [can also be found here](https://docs.nersc.gov/filesystems/htar/#htar-usage-examples).
+
+**Create an archive with a directory and file**
+
+```
+nersc$ htar -cvf archive.tar project_directory some_extra_file.json
+```
+**List the contents of a `tar` archive**
+```
+nersc$ htar -tf archive.tar
+HTAR: drwx------  als/als          0 2010-09-24 14:24  project_directory/cool_scan1
+HTAR: -rwx------  als/als    9331200 2010-09-24 14:24  project_directory/cool_scan2
+HTAR: -rwx------  als/als    9331200 2010-09-24 14:24  project_directory/cool_scan3
+HTAR: -rwx------  als/als    9331200 2010-09-24 14:24  project_directory/cool_scan4
+HTAR: -rwx------  als/als     398552 2010-09-24 17:35  some_extra_file.json
+HTAR: HTAR SUCCESSFUL
+
+```
+
+**Extract the entire `htar` file**
+
+```
+htar -xvf archive.tar
+```
+
+**Extract a single file from `htar`**
+
+```
+htar -xvf archive.tar project_directory/cool_scan4
+```
+
+**`-Hnostage` option**
+
+If your `htar` files are >100GB, and you only want to extract one or two small member files, you may find faster retrieval rates by skipping staging the file to the HPSS disk cache with `-Hnostage`.
+
+```
+htar -Hnostage -xvf archive.tar project_directory/cool_scan4
+```
+
+## Transferring Data from CFS to HPSS
+
+NERSC provides a special `xfer` QOS ("Quality of Service") for interacting with HPSS, which we can use with our SFAPI Slurm job scripts.
+
+### Single Files
+
+We can transfer single files over to HPSS using `hsi put` in a Slurm script:
+
+**Example `hsi` transfer job**
+
+```
+#SBATCH --qos=xfer
+#SBATCH -C cron
+#SBATCH --time=12:00:00
+#SBATCH --job-name=my_transfer
+#SBATCH --licenses=SCRATCH
+#SBATCH --mem=20GB
+
+# Archive a user's project folder to HPSS
+hsi put /global/cfs/cdirs/als/data_mover/8.3.2/raw/als_user_project_folder/cool_scan1.h5
+```
+
+Notes:
+- `xfer` jobs specifying -N nodes will be rejected at submission time. By default, `xfer` jobs get 2GB of memory allocated. The memory footprint scales somewhat with the size of the file, so if you're archiving larger files, you'll need to request more memory. You can do this by adding `#SBATCH --mem=XGB` to the above script (where X in the range of 5 - 10 GB is a good starting point for large files).
+- NERSC users are at most allowed 15 concurrent `xfer` sessions, which can be used strategically for parallel transfers and reads.
+
+
+### Multiple Files
+
+NERSC recommends that when serving many files smaller than 100 GB we use `htar` to bundle them together before archiving. Since individual scans within a project may not be this large, we try to archive all of the scans in a project into a single `tar` file. If projects end up being larger than 2 TB, we can create multiple `tar` files.
+
+One great part about `htar` is that it builds the archive directly on `HPSS`, so you do not need to worry about needing the additional storage allocation on the CFS side for the `tar` file.
+
+**Example `xfer` transfer job**
+```
+#SBATCH --qos=xfer
+#SBATCH -C cron
+#SBATCH --time=12:00:00
+#SBATCH --job-name=my_transfer
+#SBATCH --licenses=SCRATCH
+#SBATCH --mem=100GB
+
+# Archive a user's project folder to HPSS
+htar -cvf als_user_project.tar /global/cfs/cdirs/als/data_mover/8.3.2/raw/als_user_project_folder
+```
+
+## Transferring Data from HPSS to CFS
+
+At some point you may want to access data from HPSS. An important thing to consider is whether you need to access single or multiple files.
+
+You could extract an entire `htar` file
+
+```
+htar -xvf als_user_project_folder.tar
+```
+
+Or maybe a single file 
+
+```
+htar -xvf als_user_project_folder.tar cool_scan1.h5
+```
+
+## Prefect Flows for HPSS Transfers
+
+Most of the time we expect transfers to occur from CFS to HPSS on a scheduled basis, after users have completed scanning during their alotted beamtime.
+
+### Transfer to HPSS Implementation
+**`orchestration/transfer_controller.py`: `CFSToHPSSTransferController()`**
+
+Input
+Output
+
+### Transfer to CFS Implementation
+**`orchestration/transfer_controller.py`: `HPSSToCFSTransferController()`**
+
+Input
+Output
+
+## Update SciCat with HPSS file paths
+
+TBD
+
+
--- a/orchestration/_tests/test_sfapi_flow.py
+++ b/orchestration/_tests/test_sfapi_flow.py
@@ -37,9 +37,13 @@ def test_create_sfapi_client_success():
     """
     Test successful creation of the SFAPI client.
     """
-    from orchestration.flows.bl832.nersc import NERSCTomographyHPCController
+    from orchestration.sfapi import create_sfapi_client
 
-    # Mock data for client_id and client_secret files
+    # Define fake credential file paths
+    fake_client_id_path = "/path/to/client_id"
+    fake_client_secret_path = "/path/to/client_secret"
+
+    # Mock file contents
     mock_client_id = 'value'
     mock_client_secret = '{"key": "value"}'
 
@@ -47,34 +51,26 @@ def test_create_sfapi_client_success():
     mock_open_client_id = mock_open(read_data=mock_client_id)
     mock_open_client_secret = mock_open(read_data=mock_client_secret)
 
-    with patch("orchestration.flows.bl832.nersc.os.getenv") as mock_getenv, \
-         patch("orchestration.flows.bl832.nersc.os.path.isfile") as mock_isfile, \
+    with patch("orchestration.sfapi.os.path.isfile") as mock_isfile, \
          patch("builtins.open", side_effect=[
              mock_open_client_id.return_value,
              mock_open_client_secret.return_value
          ]), \
-            patch("orchestration.flows.bl832.nersc.JsonWebKey.import_key") as mock_import_key, \
-            patch("orchestration.flows.bl832.nersc.Client") as MockClient:
+            patch("orchestration.sfapi.JsonWebKey.import_key") as mock_import_key, \
+            patch("orchestration.sfapi.Client") as MockClient:
 
-        # Mock environment variables
-        mock_getenv.side_effect = lambda x: {
-            "PATH_NERSC_CLIENT_ID": "/path/to/client_id",
-            "PATH_NERSC_PRI_KEY": "/path/to/client_secret"
-        }.get(x, None)
-
-        # Mock file existence
+        # Simulate that both credential files exist
         mock_isfile.return_value = True
 
-        # Mock JsonWebKey.import_key to return a mock secret
+        # Mock key import to return a fake secret
         mock_import_key.return_value = "mock_secret"
 
-        # Create the client
-        client = NERSCTomographyHPCController.create_sfapi_client()
+        # Create the client using the provided fake paths
+        client = create_sfapi_client(fake_client_id_path, fake_client_secret_path)
 
-        # Assert that Client was instantiated with 'value' and 'mock_secret'
+        # Verify that Client was instantiated with the expected arguments
         MockClient.assert_called_once_with("value", "mock_secret")
-
-        # Assert that the returned client is the mocked client
+        # Assert that the returned client is the mocked Client instance
         assert client == MockClient.return_value, "Client should be the mocked sfapi_client.Client instance"
 
 
@@ -82,36 +78,25 @@ def test_create_sfapi_client_missing_paths():
     """
     Test creation of the SFAPI client with missing credential paths.
     """
-    from orchestration.flows.bl832.nersc import NERSCTomographyHPCController
+    from orchestration.sfapi import create_sfapi_client
 
-    with patch("orchestration.flows.bl832.nersc.os.getenv", return_value=None):
-        with pytest.raises(ValueError, match="Missing NERSC credentials paths."):
-            NERSCTomographyHPCController.create_sfapi_client()
+    # Passing None for both paths should trigger a ValueError.
+    with pytest.raises(ValueError, match="Missing NERSC credentials paths."):
+        create_sfapi_client(None, None)
 
 
 def test_create_sfapi_client_missing_files():
     """
     Test creation of the SFAPI client with missing credential files.
     """
-    with (
-        # Mock environment variables
-        patch(
-            "orchestration.flows.bl832.nersc.os.getenv",
-            side_effect=lambda x: {
-                "PATH_NERSC_CLIENT_ID": "/path/to/client_id",
-                "PATH_NERSC_PRI_KEY": "/path/to/client_secret"
-            }.get(x, None)
-        ),
+    from orchestration.sfapi import create_sfapi_client
+    fake_client_id_path = "/path/to/client_id"
+    fake_client_secret_path = "/path/to/client_secret"
 
-        # Mock file existence to simulate missing files
-        patch("orchestration.flows.bl832.nersc.os.path.isfile", return_value=False)
-    ):
-        # Import the module after applying patches to ensure mocks are in place
-        from orchestration.flows.bl832.nersc import NERSCTomographyHPCController
-
-        # Expect a FileNotFoundError due to missing credential files
+    # Simulate missing credential files by patching os.path.isfile to return False.
+    with patch("orchestration.sfapi.os.path.isfile", return_value=False):
         with pytest.raises(FileNotFoundError, match="NERSC credential files are missing."):
-            NERSCTomographyHPCController.create_sfapi_client()
+            create_sfapi_client(fake_client_id_path, fake_client_secret_path)
 
 # ----------------------------
 # Fixture for Mocking SFAPI Client
--- a/orchestration/sfapi.py
+++ b/orchestration/sfapi.py
@@ -1,0 +1,46 @@
+from dotenv import load_dotenv
+import json
+import logging
+import os
+
+from authlib.jose import JsonWebKey
+from sfapi_client import Client
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+load_dotenv()
+
+
+def create_sfapi_client(
+    client_id_path: str,
+    client_secret_path: str,
+) -> Client:
+    """Create and return an NERSC client instance"""
+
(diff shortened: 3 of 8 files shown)


davramov:
# Summary of Issue 59 Changes

## Key New Features

-   **New Prefect HPSS Workflows**:
	-  Added dispatcher flows to coordinate a few modes
		-   `archive_832_project_dispatcher` flow for individual project archiving
	    -   `archive_832_projects_from_previous_cycle_dispatcher` flow that implements a bi-annual archiving policy:
	        -   January 2: Archives projects from January 1 to July 15 of previous year
	        -   July 4: Archives projects from July 16 to December 31 of previous year
	-  Added a new module (`orchestration/hpss.py`) for tape archive operations at NERSC
	-   Implemented transfer controllers for CFS (Community File System) to HPSS and HPSS to CFS operations
		- `CFSToHPSSTransferController` that generates SLURM job scripts to bundle files by date ranges into tar archives for efficient tape storage
		-   `HPSSToCFSTransferController` with support for three retrieval modes: single file, full tar extraction, and partial tar extraction
	-   Added pruning controller specific to HPSS requirements
-   **Transfer and Prune Controller Refactoring**
    -   Implemented abstract base classes for both transfer and pruning operations
    -   Added type-safe controllers with better error handling
    -   Created a factory pattern for selecting appropriate controllers based on operation type
- **Endpoint Abstractions**
    -   Added new endpoint classes in `transfer_endpoints.py`:
        -   `TransferEndpoint` abstract base class
        -   `FileSystemEndpoint` for local file operations
        -   `HPSSEndpoint` for tape interactions
    -   Type-safe controller implementations using Python's `Generic[Endpoint]` pattern
-   **SciCat Integration Improvements**
    -   Enhanced dataset ingestor with better error handling
    -   Added support for derived datasets in addition to raw datasets
	-   Implemented tomo recon thumbnails from tiffs and zarr
-   **Configuration Management**: Implemented Dynaconf for more robust settings management
-   **Code Structure**: Reorganized flows by separating concerns into modular components
-   **Testing**: Added pytest test suite to ensure reliability
-   **Documentation**: Enhanced README with workflow status tables and detailed setup instructions

davramov:
I met with @rajasriramoju  and came up with a few final steps to test and verify the feature changes.

End-to-end tests
- Documentation how to test each feature
- Dummy data to move around, but real endpoints (if possible)
- Test each method for each controller
- Define a logical path to test each portion

Define test for tape:
- Look at code and see if the logic makes sense
- Come up with a simple test, move some of Dula's data to tape

orchestration/flows/scicat/ingestor_controller.py
```diff
@@ -0,0 +1,328 @@
...
+    ) -> str:
+        """Ingest data from the beamline.
+
+        :param file_path: Path to the file to ingest.
+        :return: SciCat ID of the dataset.
+        """
+        pass
+
+    @abstractmethod
+    def ingest_new_derived_dataset(
```
dylanmcreynolds, line 114: raw_dataset_id should be optional.

orchestration/flows/scicat/ingestor_controller.py
```diff
@@ -0,0 +1,328 @@
...
+        :param file_path: Path to the file to ingest.
+        :return: SciCat ID of the dataset.
+        """
+        pass
+
+    def add_new_dataset_location(
+        self,
+        dataset_id: Optional[str] = None,
+        proposal_id: Optional[str] = None,
+        file_name: Optional[str] = None,
```
dylanmcreynolds, line 130: There are three optional parameters, but you need either dataset_id or proposal_id+file_name. I think it would be cleaner if this method required dataset_id, and the find dataset were more of a utility funciton for those use cases where it might not be known.

dylanmcreynolds (commented):
A couple preliminary requests.

As discussed, let's move a lot of this stuff into the scicat_beamline project.

docs/mkdocs/docs/tomography_workflow.md
rajasriramoju, line 1: new_file_832 prefect flow is kicked off by the dispatcher, but also from spot832?

docs/mkdocs/docs/hpss.md
rajasriramoju, line 1: 2 questions:
- I expect the BL scientist or prefect dispatcher to start off the hpss transfer, because they would have access. So when is a user, mentioned in the docs, involved in this? 
- When talking about transferring specific user data from hpss from cfs, where is the operation taking place from and who is performing it?

orchestration/flows/scicat/utils.py
```diff
@@ -63,32 +102,36 @@ def calculate_access_controls(username, beamline, proposal) -> Dict:
...
-    description = [term.lower() for term in terms if len(term) > 0]
-    return " ".join(description)
+def clean_email(email: str):
+    """Clean up email addresses."""
+
+    if email:
+        if not email or email.upper() == "NONE":
+            # this is a brutal case, but the beamline sometimes puts in "None" and
+            # the new scicat backend hates that.
+            unknown_email = "unknown@example.com"
```
rajasriramoju, line 112: Can this unknown email cause potential errors down the line?

docs/mkdocs/docs/tomography_workflow.md
davramov: I see the confusion. Some arrows represent jobs getting kicked off (i.e. spot832 filewatcher -> dispatcher), but then others are data transfer, like the arrow from spot832 to the new_file_832 flow. Hmm I can adjust the diagram to be clearer what's happening within the new_file_832 flow

docs/mkdocs/docs/hpss.md
davramov: 1. 
I'm updating the documentation to define what we mean by NERSC user, ALS user, and service accounts. 

NERSC users can move data to HPSS.
In this case, "user" is assumed to be our `alsdev` service account
In production, ALS users would not be moving the data themselves, but they would still have access to their own data and know where it is located on HPSS via SciCat. As ALS staff, we could get their data if they request it using our flows, or the user could follow the documentation and log in to HPSS in jupyter and they should have file permissions for their projects.

Also thinking maybe we should set the nersc account in the slurm job as an env variable so it's easier for other people to use it themselves. Maybe a fallback can be set in `config.yml`, and overridden by the env variable.

Change it from this
`#SBATCH -A als                            # Specify the account.`

to something like this

```python
nersc_account = os.environ.get("NERSC_ACCOUNT"),
job_script = rf"""#!/bin/bash
...
#SBATCH -A {nersc_account}                            # Specify the account.
```


2. 
- The flow is set up to use the `alsdev` service account handles the transfer using the `xfer` QOS: `#SBATCH -q xfer` to launch the slurm job with the specific `hsi` and `htar` commands
- It would be launched via a Prefect Agent running on the beamline's respective VM (i.e. `flow-prd`). We have options if we want that to be a scheduled transfer, or manual transfer of one project directory.
- ALS Users who are NERSC users could look up their data in SciCat and get their data via hsi/htar commands

orchestration/flows/scicat/utils.py
davramov: Hmm, I guess we could consider cleaning it at some point down the line. 
SciCat requires that there is an email or otherwise ingestion fails. My understanding is sometimes the proposal info isn't fully pulled by labview when a scan is taken, or it could be malformed.
Maybe we could set the email to at least be the filename, like `f"unknown_email@{filename}.com"` so it's unique?

davramov: Maybe this PR is a better place to consider this: https://github.com/als-computing/splash_flows_globus/pull/50

davramov:
Notes from Raja
- what to `env` variables to be set
- add comments in the code what will move to `beamline_scicat` repo
- credentials to use for the end-to-end tests
    - scicatlive, ingest user, api url password
    - globus
- als users / HPSS tape
    - how do we expect users to access their data
    - do we need a form, clear guide, to teach beamline scientists?
- small hackathon next week to test

requirements.txt
rajasriramoju, line 1: requirements.txt is missing pytest_mock

davramov:
Notes from today's testing hackathon with Raja, just a few items to clean up

TODO:
- Write a script to ls a tar on hpss with alsdev (or whatever account moved the data)
- Within the slurm scripts (or when it determines the bundles) we might need to filter for files greater than 60 GB we use hsi instead of htar. That would move just the large files individually, and would not be in the .tar archive. Also update the logs to reflect where those files are
- In the logs, it does not print the HPSS tree, despite saying it is!
- For HPSS to CFS, make sure that the logs are not saved in a complicated tree. Maybe convert slashes to underscores for the log file name to make it easier to find?
- Update documentation on the root readme

davramov:
Hi @GBirkel, adding you as a reviewer to this PR, but don't worry about it until I rebase with other recent changes.

docs/mkdocs/docs/hpss.md
```diff
@@ -0,0 +1,410 @@
...
+   2. Create the destination directory on HPSS if it doesn't exist.
+      - Recursively check each part of the incoming file path if the folder exists
+      - If the folder does not exist, use `hsi mkdir`
+      - Repeat until the file path is built
+   3. Determine if the source is a file or a directory.
+      - If a file, transfer it using 'hsi cput'.
+      - If a directory, group files by beam cycle and archive them.
+         * Cycle 1: Jan 1 - Jul 15
+         * Cycle 2: Jul 16 - Dec 31
+         * If a group exceeds 2 TB, it is partitioned into multiple tar archives.
```
GBirkel, line 302: Should there be code that warns if the resulting archive would be less than 100GB, and skips it, so the content is consolidated into the next archiving run instead?

orchestration/flows/scicat/ingestor_controller.py
```diff
@@ -0,0 +1,324 @@
...
+
+        :param file_path: Path to the file to ingest.
+        :return: SciCat ID of the dataset.
+        """
+        pass
+
+    def add_new_dataset_location(
+        self,
+        dataset_id: str = None,
+        source_folder: str = None,
```
GBirkel, line 131: Why are we giving `add_new_dataset_location` a path to a folder, then having it construct a file path by appending the name of the data set to the folder?  It seems like a dangerous assumption to make that the file would always have that name, especially since it appears we are quietly falling back to a generic name of "dataset".
If a users wants to put two data sets in the same place and renames one for this purpose, there is no way to supply the changed name to `add_new_dataset_location`.
Perhaps we should leave it to the caller to compose the full filename?

orchestration/flows/scicat/ingestor_controller.py
```diff
@@ -0,0 +1,324 @@
...
+
+            # Create a minimal datablock for the new location
+            datablock = CreateDatasetOrigDatablockDto(
+                size=size,
+                dataFileList=[datafile]
+            )
+
+            # Add location information to the path if host is provided
+            if source_folder_host:
+                datafile.path = f"{source_folder_host}:{file_path}"
```
GBirkel, line 176: Perhaps prepare this string before creating the `DataFile` object above, so we're not modifying the object after creation?

orchestration/flows/scicat/ingestor_controller.py
```diff
@@ -0,0 +1,324 @@
...
+            else:
+                logger.error(f"HTTP error occurred: {e}")
+        except Exception as e:
+            logger.error(f"Failed to remove datablock from dataset {dataset_id}: {e}")
+        return False
+
+    def _find_dataset(
+        self,
+        proposal_id: Optional[str] = None,  # The ALS proposal ID, not the SciCat ID
+        file_name: Optional[str] = None
```
GBirkel, line 254: I'm not sure if SciCat's advanced query API supports this, but, if we're actually searching by `file_name`, wouldn't it make more sense to look in all the dataFileList entries for all datasets?

This comes to mind because at 733, scientists organize their data mostly by creating dated folders, and there's no guarantee that the files in those folders have unique names relative to the other folders.  So if they were searching for a particular data file, they would need to use a path fragment, e.g. '20241216_153047/new_run.h5'

If this function could search dataFileLists by path fragment, it would be some future-proofing for those users...

orchestration/flows/scicat/utils.py
```diff
@@ -40,7 +47,39 @@ def default(self, obj):
         return json.JSONEncoder.default(self, obj)
 
 
-def calculate_access_controls(username, beamline, proposal) -> Dict:
+def build_search_terms(
+    sample_name: str
+) -> str:
+    """extract search terms from sample name to provide something pleasing to search on"""
+    terms = re.split("[^a-zA-Z0-9]", sample_name)
```
GBirkel, line 54: Should this include dashes or underscores?

orchestration/prune_controller.py
```diff
@@ -0,0 +1,390 @@
...
+                check_endpoint=check_endpoint,
+                config=self.config
+            )
+        else:
+            # Otherwise, schedule pruning for future execution
+            logger.info(f"Scheduling pruning of '{file_path}' from '{source_endpoint.name}' "
+                        f"in {days_from_now.total_seconds()/86400:.1f} days")
+
+            try:
+                schedule_prefect_flow(
```
GBirkel, line 142: What happens if this is called twice, and two prining jobs for the same location are triggered at the same time?  Will we create a race between the jobs and end up with file errors?

orchestration/prune_controller.py
```diff
@@ -0,0 +1,390 @@
...
+        """
+        super().__init__(config)
+        logger.debug(f"Initialized FileSystemPruneController for beamline {config.beamline_id}")
+
+    def prune(
+        self,
+        file_path: str = None,
+        source_endpoint: FileSystemEndpoint = None,
+        check_endpoint: Optional[FileSystemEndpoint] = None,
+        days_from_now: datetime.timedelta = 0
```
GBirkel, line 99: This is asking for `days_from_now` but expects a `timedelta`.  Might be more clear to either:
1. Change the argument to `time_delta`
2. Expect a floating point value, and create the `timedelta` internally from it

orchestration/prune_controller.py
```diff
@@ -0,0 +1,390 @@
...
+                config=self.config
+            )
+        else:
+            # Otherwise, schedule pruning for future execution
+            logger.info(f"Scheduling pruning of '{file_path}' from '{source_endpoint.name}' "
+                        f"in {days_from_now.total_seconds()/86400:.1f} days")
+
+            try:
+                schedule_prefect_flow(
+                    deployment_name="prune_filesystem_endpoint/prune_filesystem_endpoint",
```
GBirkel, line 286: Should this be `prune_globus_endpoint` here?

orchestration/sfapi.py
```diff
@@ -0,0 +1,45 @@
...
+    client_secret_path: str = os.getenv("PATH_NERSC_PRI_KEY"),
+) -> Client:
+    """Create and return an NERSC client instance"""
+
+    # When generating the SFAPI Key in Iris, make sure to select "asldev" as the user!
+    # Otherwise, the key will not have the necessary permissions to access the data.
+
+    if not client_id_path or not client_secret_path:
+        logger.error("NERSC credentials paths are missing.")
+        raise ValueError("Missing NERSC credentials paths.")
```
GBirkel, line 26:
```suggestion
        raise ValueError("NERSC credentials paths are missing.")
```

orchestration/prune_controller.py
davramov: From what I've seen, if the file is already pruned, then the flow would fail with a "file not found" error

orchestration/flows/scicat/utils.py
```diff
@@ -40,7 +47,39 @@ def default(self, obj):
...
+    return " ".join(description)
+
+
+def build_thumbnail(
+    image_array: npt.ArrayLike
+) -> io.BytesIO:
+    """Create a thumbnail from an image array."""
+    image_array = image_array - np.min(image_array) + 1.001
+    image_array = np.log(image_array)
+    image_array = 205 * image_array / (np.max(image_array))
```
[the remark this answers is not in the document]
davramov: I'm adding a try-except statement, where if an error occurs while building the thumbnail (e.g., dividing by 0), it will return a blank image. This way if it fails, it won't block the rest of the ingestion flow from completing.

docs/mkdocs/docs/hpss.md
```diff
@@ -0,0 +1,410 @@
...
+
+#### Transfer to HPSS Implementation
+**`orchestration/transfer_controller.py`:**
+  - **`CFSToHPSSTransferController()`**: This controller uses a Slurm Job Script and SFAPI to launch the tape transfer job. The Slurm script handles the specific logic for handling single and multiple files, on a project by project basis. It reads the files sizes, and creates bundles that are <= 2TB. The groups within each tar archive are saved in a log on NERSC CFS for posterity.
+
+  Here is a high level overview of the steps taken within the SFAPI Slurm Job:
+   1. Define the source (CFS) and destination (HPSS) paths.
+   2. Create the destination directory on HPSS if it doesn't exist.
+      - Recursively check each part of the incoming file path if the folder exists
+      - If the folder does not exist, use `hsi mkdir`
```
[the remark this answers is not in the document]
davramov: Good catch. I found this in the documentation: https://hpss-collaboration.org/wp-content/uploads/2023/09/hpss_hsi_10.2_reference_manual.pdf
```
19.63. MKDIR command
Synopsis
mkd[ir] [-A annotation] [-m mode] [-p] path …
Description
This command creates HPSS subdirectories.
HSI Version
All
Aliases
add, md
Options
...
-p Creates missing intermediate path name directories. If the -p flag is not specified,
the parent directory of each newly-created directory must already exist
```

docs/mkdocs/docs/hpss.md
davramov: Interesting thought. The lower limit bound (<100GB) seems more like a suggestion so that small files are not scattered across many different tapes. I think if all of the files are bundled together for a cycle/proposal, then users could still retrieve their data easily.

orchestration/transfer_controller.py
```diff
@@ -115,18 +63,29 @@ def copy(
...
+
     def __init__(
         self,
-        config: Config832,
+        config: BeamlineConfig,
         prometheus_metrics: Optional[PrometheusMetrics] = None
     ) -> None:
         super().__init__(config)
         self.prometheus_metrics = prometheus_metrics
+        logger.debug(f"Initialized GlobusTransferController for beamline {config.beamline_id}")
```
dylanmcreynolds, line 82: Since you're calling the `super().__init__`, I would assume you're getting this log twice. Since the super is logging yhr class, this is extraneous, no?

orchestration/_tests/test_prune_controller.py
```diff
@@ -0,0 +1,203 @@
...
+        Dummy implementation of the abstract method.
+        (Not used in these tests.)
+        """
+        return True
+
+
+###############################################################################
+# Pytest Fixtures
+###############################################################################
+@pytest.fixture
```
dylanmcreynolds, line 95: pytest already has this fixture: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture,probably use it.

orchestration/_tests/test_prune_controller.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    Fixture that returns an instance of DummyPruneController using the temporary directory.
+    """
+    return DummyPruneController(base_dir=test_dir)
+
+
+###############################################################################
+# Helper Function for Creating Test Files
+###############################################################################
+def create_test_files(directory: Path, dates):
```
dylanmcreynolds, line 118: This could be a fixture, no?

orchestration/_tests/test_prune_controller.py
```diff
@@ -0,0 +1,203 @@
...
+import pytest
+
+# Import the abstract PruneController base
+from orchestration.prune_controller import PruneController
+
+
+###############################################################################
+# Dummy Implementation for Testing
+###############################################################################
+class DummyPruneController(PruneController):
```
dylanmcreynolds, line 15: "Mock" is a more standard term than "Dummy"

orchestration/_tests/test_prune_controller.py
```diff
@@ -0,0 +1,203 @@
...
+    from orchestration.prune_controller import PruneController
+    # Verify that our dummy controller is an instance of the abstract base class
+    assert isinstance(prune_controller, PruneController)
+    # And that the base directory exists
+    assert prune_controller.base_dir.exists()
+
+
+def test_get_files_to_delete(prune_controller, test_dir):
+    # Create test files with various modification times.
+    now = datetime.now()
```
dylanmcreynolds, line 151: It seems to me that that this test is testing the `DummyPruneController` more than actual code, no?

orchestration/_tests/test_prune_controller.py
```diff
@@ -0,0 +1,203 @@
...
+        Args:
+            retention (int): Retention period in days. Must be > 0.
+
+        Returns:
+            List[Path]: Files older than the retention period.
+
+        Raises:
+            ValueError: If retention is not positive.
+        """
+        if retention <= 0:
```
dylanmcreynolds, line 46: Should this logic be in one of the real instances?

orchestration/_tests/test_sfapi_flow.py
```diff
@@ -37,81 +37,66 @@ def test_create_sfapi_client_success():
...
-            patch("orchestration.flows.bl832.nersc.JsonWebKey.import_key") as mock_import_key, \
-            patch("orchestration.flows.bl832.nersc.Client") as MockClient:
-
-        # Mock environment variables
-        mock_getenv.side_effect = lambda x: {
-            "PATH_NERSC_CLIENT_ID": "/path/to/client_id",
-            "PATH_NERSC_PRI_KEY": "/path/to/client_secret"
-        }.get(x, None)
+            patch("orchestration.sfapi.JsonWebKey.import_key") as mock_import_key, \
+            patch("orchestration.sfapi.Client") as MockClient:
```
dylanmcreynolds, line 60: Super picky, but this should be snake_case

dylanmcreynolds (changes requested):
I did a partial review, with a few changes. I'll do more later.

orchestration/flows/bl832/dispatcher.py
```diff
@@ -158,6 +177,213 @@ async def dispatcher(
...
+
+    # ARCHIVE ALL PROJECTS IN THE NERSC / CFS / 8.3.2 / SCRATCH DIRECTORY
+    for project in config.tc.operation_ls(
+        endpoint_id=config.nersc832.endpoint_id,
+        path=config.nersc832_alsdev_scratch.path,
+        orderby=["name", "last_modified"],
+    ):
+        logger.info(f"Found project: {project}")
+        try:
+            run_specific_flow(
```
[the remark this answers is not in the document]
davramov: I added inline comments around these blocks

orchestration/transfer_endpoints.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Args:
+            path_suffix (str): The relative path to append.
+
+        Returns:
+            str: The full absolute path.
+        """
+        if path_suffix.startswith("/"):
+            path_suffix = path_suffix[1:]
+        return f"{self.root_path.rstrip('/')}/{path_suffix}"
```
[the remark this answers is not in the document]
davramov: Updated to use Pathlib:

`return str(Path(self.root_path) / path_suffix)`

orchestration/_tests/test_transfer_controller.py
```diff
@@ -299,53 +317,361 @@ def test_simple_transfer_controller_no_source_or_destination(mock_config832, tra
...
+#             controller = SimpleTransferController(mock_config832)
+#             result = controller.copy(
+#                 file_path="some_dir/test_file.txt",
+#                 source=mock_file_system_endpoint,
+#                 destination=mock_file_system_endpoint,
+#             )
+#             assert result is False, "Expected False when os.system returns non-zero."
+#             mock_os_system.assert_called_once()
+#             command_called = mock_os_system.call_args[0][0]
+#             assert "cp -r" in command_called, "Expected cp command in os.system call."
```
kkamdin: if you're happy with the new version of the test we should remove this commented out test before merging

orchestration/flows/bl832/move.py
```diff
@@ -117,100 +46,126 @@ def process_new_832_file(file_path: str,
...
+    logger.info(f"File successfully transferred from data832 to NERSC {new_file}. Success: {nersc_success}")
+    pass
 
-@flow(name="test_832_transfers_grafana")
-def test_transfers_832_grafana(file_path: str = "/raw/transfer_tests/test/"):
-    logger = get_run_logger()
-    config = Config832()
 
-    task = transfer_data_to_nersc(file_path, config.tc, config.data832, config.nersc_alsdev)
+if __name__ == "__main__":
```
kkamdin: is this main function for testing or used to run the flow?

orchestration/_tests/test_prune_controller.py
```diff
@@ -0,0 +1,320 @@
...
+    """If scheduling fails, fs_controller.prune should return False."""
+    rel = "error.txt"
+    create_file_or_dir(tmp_fs_path, rel)
+
+    assert fs_controller.prune(
+        file_path=rel,
+        source_endpoint=fs_endpoint,
+        check_endpoint=None,
+        days_from_now=2.0,
+    ) is False
```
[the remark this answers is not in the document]
davramov: I have a negative days case now in pytest

orchestration/flows/scicat/utils.py
```diff
@@ -40,7 +47,53 @@ def default(self, obj):
...
+    terms = re.split("[^a-zA-Z0-9]", sample_name)
+    description = [term.lower() for term in terms if len(term) > 0]
+    return " ".join(description)
+
+
+def build_thumbnail(
+    image_array: npt.ArrayLike
+) -> io.BytesIO:
+    """Create a thumbnail from an image array."""
+    try:
```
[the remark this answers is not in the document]
davramov: I like these SciCat suggestions, and I created a new issue to track them separately: https://github.com/als-computing/splash_flows/issues/97

orchestration/flows/bl832/scicat_ingestor.py
```diff
@@ -0,0 +1,764 @@
...
+        beamline,
+        proposal
+    ) -> Dict:
+        """
+        Calculate access controls for a dataset.
+        """
+
+        # make an access group list that includes the name of the proposal and the name of the beamline
+        access_groups = []
+        # sometimes the beamline name is super dirty  " '8.3.2', "" '8.3.2', "
```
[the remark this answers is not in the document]
davramov: Let's track this in the new SciCat issue: https://github.com/als-computing/splash_flows/issues/97

orchestration/flows/scicat/ingestor_controller.py
```diff
@@ -0,0 +1,328 @@
...
+            response = self.scicat_client.datasets_delete(datablock_id)
+            if response:
+                logger.info(f"Successfully removed datablock {datablock_id} from dataset {dataset_id}")
+                return True
+            else:
+                logger.error(f"Failed to delete datablock {datablock_id} from dataset {dataset_id}")
+                return False
+
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code == 403:
```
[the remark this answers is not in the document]
davramov: SciCat clients can be one of a few groups: `["admin","ingestor","archivemanager"]`, and it looks like `archivemanager` manager would be the one to be able to delete datasets. I'll track this in the new SciCat issue https://github.com/als-computing/splash_flows/issues/97 rather than this PR.

PR #71, opened 2025-04-07
Xiaoya new feature grafana

xiaoyachong:
This PR adds functionality to collect and visualize data transfer metrics via **Prometheus** and **Grafana**. Key changes include:

1. Metrics Collection and Prometheus Integration:
- Added a new `prometheus_utils.py` module for pushing metrics to Prometheus Pushgateway
- Implemented metrics for transfer counts, file sizes, transfer speeds, and durations

2. Enhanced Transfer Controller:
- Expanded the `GlobusTransferController` with file size detection and metrics collection
- Modified the copy method to support metrics collection with optional flags

3. New Flow for Grafana Testing:
- Added `test_transfers_832_grafana` flow to test the metrics collection
- Created `transfer_data_to_ners_by_controller` task that uses the enhanced controller
- Added a deployment script `create_deployments_832_grafana.sh`

4. Configuration Updates:
- Modified NERSC endpoint path in `config.yml` to use a test directory
- Added Prometheus client dependency to `requirements.txt`

---
  - https://app.asana.com/0/0/1209588090194771


orchestration/prometheus_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    JOB_NAME = os.getenv('JOB_NAME', 'nersc_transfer')
+    INSTANCE_LABEL = os.getenv('INSTANCE_LABEL', 'data_transfer')
+    
+    try:
+        # Create a new registry
+        registry = CollectorRegistry()
+        
+        # Define the required metrics
+        # 1. Count of requests - Gauge
+        request_counter = Gauge('nersc_transfer_request_count',
```
taxe10, line 18: Currently, I believe that this set of `Gauge`(s) are created/defined every time we call `push_metrics_to_prometheus`. Is it feasible to define them once and just update them when calling this function. Perhaps something like:
```
class PrometheusMetrics():
  def __init__(self):
    try:
      self.registry = CollectorRegistry()
      self.request_counter = Gauge('nersc_transfer_request_count', ....
      ...

  def push_metrics_to_prometheus(self, metrics, logger):
     try:
            execution_id = f"exec_{uuid.uuid4()}"
            ...

            # Set metric values
            self.request_counter.labels(execution_id=execution_id).set(1)
            ...
```

orchestration/transfer_controller.py
```diff
@@ -95,6 +97,8 @@ def copy(
         file_path: str = None,
         source: Endpoint = None,
         destination: Endpoint = None,
+        collect_metrics: bool = True,
+        machine_name: str = "default"
```
taxe10, line 101: Should the `machine_name` be validated against a set of pre-defined options (e.g. NERSC, ALCF, etc)?
Maybe something like:
```
class MachineName(Enum):
    NERSC = "nersc"
    ALCF = "alcf"
    ...
```

Food for thought - and perhaps something to consider for a future PR, is that we may need to track metrics not only by machine, but by endpoint as well? @dylanmcreynolds, what do you think?

requirements.txt
```diff
@@ -13,4 +13,5 @@ pyyaml
 authlib
 sfapi_client
 globus-compute-sdk @ git+https://github.com/globus/globus-compute.git@d1731340074be56861ec91d732bdff44f8e2b46e#subdirectory=compute_sdk
-griffe==0.47
\ No newline at end of file
+griffe==0.47
+prometheus_client
```
taxe10, line 17: Do we need to pin the version of prometheus?

orchestration/transfer_controller.py
```diff
@@ -111,22 +117,199 @@ def copy(
...
+            
+            # Get the full Globus path
+            globus_path = self._get_full_globus_path(path, endpoint)
+            
+            # Try to get item details - first see if it's a file or directory
+            try:
+                # First try direct stat if endpoint supports it
+                item_info = transfer_client.operation_stat(endpoint.uuid, path=globus_path)
+                is_dir = item_info['type'] == 'dir'
+            except (globus_sdk.GlobusAPIError, KeyError):
```
taxe10, line 166: Could you please offer some context on the two instances of `globus_sdk.GlobusAPIError`? Is it that some endpoints do not support some of these operations?

create_deployments_832_grafana.sh
taxe10, line 1: This looks good. I'm wondering if it may be good to integrate this in `create_deployments_832.sh`, perhaps with a flag to test grafana?

taxe10 (changes requested):
Very nice work, @xiaoyachong! I left a couple of comments for your revision and a couple of questions as well. Something we should think about is that we may need to track metrics per endpoint (or maybe a set of endpoints) in the future instead of per machine (e.g. NERSC) - which can be addressed in a future PR of course, but good to keep in mind now to avoid a major refactor later.
I'd be happy to take another look once changes have been updated.

orchestration/transfer_controller.py
davramov: In `orchestration/flows/bl832/job_controller.py` I have an enum defined for the systems we support

https://github.com/als-computing/splash_flows_globus/blob/745887edf41718f4c58071c5d19265e6faaf735d/orchestration/flows/bl832/job_controller.py#L52

```python
class HPC(Enum):
    """
    Enum representing different HPC environments.
    Use enum names as strings to identify HPC sites, ensuring a standard set of values.

    Members:
        ALCF: Argonne Leadership Computing Facility
        NERSC: National Energy Research Scientific Computing Center
    """
    ALCF = "ALCF"
    NERSC = "NERSC"
    OLCF = "OLCF"
```

config.yml
```diff
@@ -49,7 +49,7 @@ globus:
       name: alcf_home832
 
     nersc_alsdev:
-      root_path: /global/homes/a/alsdev/test_directory/
+      root_path: /global/homes/a/alsdev/test_directory_xchong/
```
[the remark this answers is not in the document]
xiaoyachong: I've reverted this change and am now using /global/homes/a/alsdev/test_directory/ instead.

orchestration/flows/bl832/move.py
```diff
@@ -192,3 +229,17 @@ def test_transfers_832(file_path: str = "/raw/transfer_tests/test.txt"):
...
         f"File successfully transferred from data832 to NERSC {spot832_path}. Task {task}"
     )
+
+
+@flow(name="test_832_transfers_grafana")
+def test_transfers_832_grafana(file_path: str = "/raw/transfer_tests/test_xchong/"):
+    logger = get_run_logger()
+    config = Config832()
+
+    logger.info(f"{str(uuid.uuid4())}{file_path}")
```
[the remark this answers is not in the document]
xiaoyachong: This logger.info() line was unused and has been removed.

orchestration/flows/bl832/move.py
```diff
@@ -74,6 +74,43 @@ def transfer_data_to_nersc(
 
     return success
 
+@task(name="transfer_data_to_nersc")
+def transfer_data_to_ners_by_controller(
```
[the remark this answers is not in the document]
xiaoyachong: I directly modified `transfer_data_to_nersc` and removed `transfer_data_to_ners_by_controller`.

orchestration/flows/bl832/move.py
```diff
@@ -192,3 +229,17 @@ def test_transfers_832(file_path: str = "/raw/transfer_tests/test.txt"):
     logger.info(
         f"File successfully transferred from data832 to NERSC {spot832_path}. Task {task}"
     )
+
+
+@flow(name="test_832_transfers_grafana")
+def test_transfers_832_grafana(file_path: str = "/raw/transfer_tests/test_xchong/"):
```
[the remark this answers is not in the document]
xiaoyachong: Thanks for your suggestion. I modified it to `/raw/transfer_tests/test/`.

orchestration/prometheus_utils.py
```diff
@@ -0,0 +1,73 @@
...
+                             registry=registry)
+        
+        # Generate a unique execution ID for this transfer
+        execution_id = f"exec_{str(uuid.uuid4())}"
+        
+        # Set the metrics
+        request_counter.labels(execution_id=execution_id).set(1)
+        bytes_counter.labels(execution_id=execution_id).set(metrics['bytes_transferred'])
+        transfer_bytes.labels(status=metrics['status']).set(metrics['bytes_transferred'])
+        transfer_time.labels(machine=metrics['status']).set(metrics['duration_seconds'])
```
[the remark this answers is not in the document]
xiaoyachong: I correct this and I'm now using `status=metrics['status']` as the label for `transfer_bytes`, `transfer_time`, and `transfer_speed`, which means the results are grouped by status. I can also switch to `machine=metrics['machine'] `if we want to group the metrics by destination instead. This label appears on the rightmost part of the dashboard—currently, with `status` as the label, `"success"` is shown under `Name`.

orchestration/prometheus_utils.py
xiaoyachong: That’s a great suggestion—I’ve defined PrometheusMetrics as you recommended.

requirements.txt
xiaoyachong: I modified it to `prometheus_client==0.21.1`

orchestration/transfer_controller.py
xiaoyachong: Thanks for your information. To make machine_name only accept values from the HPC enum, I modify the parameter type and add validation that raises an error if an invalid value is provided.

orchestration/transfer_controller.py
xiaoyachong: In my test, the endpoint I’m using supports both `operation_stat()` and `operation_ls()`. I’ve removed the `try-except` blocks and `globus_sdk.GlobusAPIError` handling to simplify the code.

create_deployments_832_grafana.sh
xiaoyachong: I integrate this with `create_deployments_832.sh`

orchestration/transfer_controller.py
```diff
@@ -111,22 +118,156 @@ def copy(
...
     """
+    def __init__(
+        self,
+        config: Config832
+    ) -> None:
+        super().__init__(config)
+        # Create metrics instance once per controller instance
+        self.prometheus_metrics = PrometheusMetrics()
+        
+    def get_file_size(
```
dylanmcreynolds, line 135: This is doing a lot of work with a fair number of messages to globus. Is there really nothing in the globus API do do this? When you start a transfer, globus definitely calculates the size of the payload. Perhaps it's available at the end of the transfer?

dylanmcreynolds (changes requested):
This is great. A few comments.

dylanmcreynolds (changes requested):
I would like to have pyunit tests for this PR before merging. Thanks!

orchestration/transfer_controller.py
```diff
@@ -111,22 +118,156 @@ def copy(
...
     Args:
         TransferController: Abstract class for transferring data.
     """
+    def __init__(
+        self,
+        config: Config832
+    ) -> None:
+        super().__init__(config)
+        # Create metrics instance once per controller instance
+        self.prometheus_metrics = PrometheusMetrics()
```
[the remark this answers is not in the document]
xiaoyachong: That sounds good. I followed your suggestion but used `self.prometheus_metrics = prometheus_metrics` (without or `PrometheusMetrics()`), so that `GlobusTransferController` supports data transfers without pushing metrics when the user doesn't provide a `prometheus_metrics` instance during initialization.

orchestration/transfer_controller.py
xiaoyachong: That’s a great point. I’ve added a new function called `get_transfer_file_info()` that retrieves task information using `transfer_client.get_task(task_id)`.

orchestration/globus/transfer.py
```diff
@@ -132,9 +133,15 @@ def start_transfer(
     # if a transfer failed, like for a file not found globus keeps trying for a long time
     # and won't let another be attempted
     task_id = task["task_id"]
-    return task_wait(
+
+    success = task_wait(
         transfer_client, task_id, max_wait_seconds=max_wait_seconds, logger=logger
     )
+
```
dylanmcreynolds, line 140: Rather than sending this boolean, why not always send it? If a client doesn't use it, they can use this call 

```
success, _ = start_transfer(....)
```

orchestration/transfer_controller.py
```diff
@@ -111,17 +114,147 @@ def copy(
...
+                "machine": machine_name
+            }
+            
+            # Push metrics to Prometheus
+            self.prometheus_metrics.push_metrics_to_prometheus(metrics, logger)
+            
+        except Exception as e:
+            logger.error(f"Error collecting or pushing metrics: {e}")
+
+    def _get_hpc_name_from_endpoint(self, endpoint: GlobusEndpoint) -> str:
```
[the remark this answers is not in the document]
xiaoyachong: I change to `machine_name = destination.name` and delete `_get_hpc_name_from_endpoint`

orchestration/globus/transfer.py
dylanmcreynolds: I don't know. I don't wnat to have to modify that much, yet having a funciton that sometimes returns one and at other returns two values is a little unsettling. I think I would choose making the change to all the files.

xiaoyachong: Thanks! I've just finished modifying all the functions mentioned above.

taxe10 (approved):

Issue #74, opened 2025-04-22
Github action - setup python


davramov:
The Github Action for the last merged PR resulted in this error, preventing any downstream configuration or pytest cases from running:

```
Run actions/setup-python@v2
  with:
    python-version: 3.11
    cache: pip
    token: ***
Successfully setup CPython (3.11.1[2](https://github.com/als-computing/splash_flows_globus/actions/runs/14600696296/job/40958632345#step:3:2))
/opt/hostedtoolcache/Python/[3](https://github.com/als-computing/splash_flows_globus/actions/runs/14600696296/job/40958632345#step:3:3).11.12/x64/bin/pip cache dir
/home/runner/.cache/pip
Error: Cache service responded with [4](https://github.com/als-computing/splash_flows_globus/actions/runs/14600696296/job/40958632345#step:3:4)22
```

I found this related issue:
https://github.com/actions/setup-python/issues/1085

They suggest bumping `actions/setup-python@v2` to `actions/setup-python@v5`

Dylan also suggested to add this to pip install: `--no-cache-dir`

davramov:
Closing this issue, it's been fixed with this PR
https://github.com/als-computing/splash_flows_globus/pull/75

PR #75, opened 2025-04-22
Issue 74: Github action setup-python cache fix

davramov:
Bumping `actions/setup-python` from v2 to v5 to solve the recent caching issue.

Ref: https://github.com/als-computing/splash_flows_globus/issues/74

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v
+    - uses: actions/checkout@v2
     - name: Set up Python 3.11
       uses: actions/setup-python@v5
       with:
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -11,15 +11,15 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/python-app.yml
```diff
@@ -12,14 +12,14 @@ jobs:
...
-      uses: actions/setup-python@v2
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v5
       with:
         python-version: 3.11
         cache: 'pip'
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
+        python -m pip install --upgrade pip --no-cache-dir
```
dylanmcreynolds, line 22: I believe the order should be 
python -m pip install  --no-cache-dir --upgrade pip

.github/workflows/python-app.yml
```diff
@@ -12,14 +12,14 @@ jobs:
...
         pip install flake8 pytest
```
dylanmcreynolds, line 23: needs the --no-cache for the following two lines as well

dylanmcreynolds (changes requested):

.github/workflows/python-app.yml
davramov: done

.github/workflows/python-app.yml
davramov: and done!

PR #84, opened 2025-08-15
Prefect 3 Upgrade

davramov:
Preparing for upgrading our Prefect Server from v2.x to v3.4.2

Other notes
- Adding "file_path" basename to flow run names for easier identification in the UI (addressing https://github.com/als-computing/splash_flows/issues/82)
- including a new files to streamline deployment with
    - `orchestration/flows/bl832/prefect.yaml` : Defines which flows from which files to deploy, and assign them to a work pool and queue
    - `init_832_work_pools.sh` : Creates work pools if they don't already exist, and deploys all of the flows defined above in `prefect.yaml`
- updating Prefect version in `requirements.txt`
- updating pytests and flows scripts to address updates within Prefect


davramov:
> Do you need to preserve the scripts in "legacy"? In general I like to keep the trail free of dead branches :)

No, I can delete these 👍

PR #86, opened 2025-09-10
9.3.1 Flows

davramov:
Preparing prefect flows for BL 9.3.1

[Asana Project](https://app.asana.com/1/4262699445529/project/1211230697574170/board/1211248581641988)

- Draft 9.3.1 globus endpoints in config.yml
- Copying boilerplate code from 7.3.3, and making it 9.3.1-specific

---
  - https://app.asana.com/0/0/1211348170629255

### DIFF
--- a/orchestration/_tests/test_bl931/test_move.py
+++ b/orchestration/_tests/test_bl931/test_move.py
@@ -108,7 +108,7 @@ def test_process_new_931_file_task(mocker: MockFixture) -> None:
     result = process_new_931_file_task(file_path=test_file_path, config=mock_config)
 
     # Verify that the transfer controller's copy method was called exactly once.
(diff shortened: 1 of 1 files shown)


davramov:
Some information from Johannes about the medium/long-term data movement strategy.

Current workflow:
1. Collect scan
2. Save the tpx3 file and the metadata file
3. convert tpx3 file to h5
4. embed metadata into h5 file

Right now, steps 2-4 run on the beamline computer. So at the moment, we just need to worry about moving a single `h5` file per scan.

However, in the future, the flow will be a little different.
Step **2** will still happen on the beamline computer, but steps **3+4** will happen on Lamar after transferring to beegfs.
This means the bl931 dispatcher will eventually need to listen for a list of files, and the move flow will need to trigger the file format conversion and metadata embedding codes on Lamar.

PR #100, opened 2025-12-11
Kate/bl7012 datamovement

kkamdin:
Tested that a source file from the new COSMIC "Phasis" DTN (/datamovement_test/test.txt) moves to NERSC destination (/global/cfs/cdirs/als/gsharing/data_mover/7012/datamovement_test/test.txt). With a new globus client that I created. This PR updates the respective uuids in the config.

Test methods:
- locally with just the globus python api & ptyhon script
- locally with prefect deployment and using the prefect UI

I updated the deployment script as well because I got errors from using `deployment build` (apparently it's too deprecated to work in my local environment? the warning said the command isn't available after September 2024).

image

Also did a visual check in the Globus UI that my test file is there on NERSC with the right time stamp:

image

### DIFF
--- a/create_deployments_7012.sh
+++ b/create_deployments_7012.sh
@@ -1,8 +1,72 @@
-export $(grep -v '^#' .env | xargs)
+#!/bin/bash
+# Create BL7012 deployment
+# Works with both local Prefect server and remote deployments
 
+set -e
+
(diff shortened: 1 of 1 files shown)


davramov:
Looks good! I'll go ahead and merge.

PR #104, opened 2025-12-23
[8.3.2] Fixing pscratch prune path for raw file

davramov:
Addressing this issue that I found: https://github.com/als-computing/splash_flows/issues/103

Now, after reconstruction is complete, the actual relative path for raw files is passed into `schedule_pruning` rather than the original absolute path.


kkamdin (approved):
LGTM! one nit that i'll leave up to you

PR #111, opened 2026-01-15
Tomography reconstruction multinode optimization

davramov:
This PR adds the `reconstruct_multinode()` method for tomography reconstruction in `orchestration/flows/bl832/nersc.py`.
- Ability to request `num_nodes` when calling reconstruction, and use the correct QOS.
- Uses shifter instead of podman to handle containers
- By partitioning sinograms across multiple CPU nodes, I achieved near-linear speedup. For 8 nodes, there was close to a 7x speedup in performance for the reconstruction (not including overhead).
- From here, I found the next main bottleneck was Podman-hpc, which pulls the microct image on every job (~90 seconds). By switching to Shifter with a pre-cached image, container startup dropped to ~2-3 seconds. The remaining overhead (~1 minute) is due to SFAPI and queuing on Perlmutter (not much we can improve here).
- The sweet spot seems to be 4 CPU nodes in the realtime queue using Shifter, bringing down the total time from ~10 minutes to ~2 minutes. This balances the quick pickup by the realtime queue and the linear performance boost. Scaling beyond this requires the regular, demand, or premium queues, which have longer wait times that offset the reconstruction speedup (maybe we can ask Bjoern nicely for more nodes in the realtime queue).
- For fun, I ran one test using 128 nodes, and while recon was fast (~30 seconds), the wait in the queue was close to 30 minutes.
image

Additionally, this PR improves the `cancel_sfapi_job.py` script, includes the reconstruction/multiresolution scripts used on Perlmutter.

### DIFF
--- a/orchestration/_tests/test_bl832/test_nersc.py
+++ b/orchestration/_tests/test_bl832/test_nersc.py
@@ -115,7 +115,7 @@ def mock_config832(mocker):
         "original_checkpoint_path": "/mock/original.pt",
         "finetuned_checkpoint_path": "/mock/checkpoints/finetuned.pt",
     }
-    mock_config.nersc_segment_dino_settings = {
+    mock_config.nersc_segment_dinov3_settings = {
         "qos": "regular",
         "account": "mock_account",
         "constraint": "gpu",
@@ -358,10 +358,10 @@ def test_segmentation_sam3_uses_variable_options(mocker, mock_sfapi_client, mock
 
 
 # ──────────────────────────────────────────────────────────────────────────────
-# segmentation_dino
+# segmentation_dinov3
 # ──────────────────────────────────────────────────────────────────────────────
 
-def test_segmentation_dino_success(mocker, mock_sfapi_client, mock_config832):
+def test_segmentation_dinov3_success(mocker, mock_sfapi_client, mock_config832):
     from orchestration.flows.bl832.nersc import NERSCTomographyHPCController
     from sfapi_client.compute import Machine
 
@@ -369,7 +369,7 @@ def test_segmentation_dino_success(mocker, mock_sfapi_client, mock_config832):
     mocker.patch("orchestration.flows.bl832.nersc.Variable.get", return_value={"defaults": True})
     controller = NERSCTomographyHPCController(client=mock_sfapi_client, config=mock_config832)
 
-    result = controller.segmentation_dino(recon_folder_path="folder/recfile")
+    result = controller.segmentation_dinov3(recon_folder_path="folder/recfile")
 
     mock_sfapi_client.compute.assert_called_with(Machine.perlmutter)
     mock_sfapi_client.compute.return_value.submit_job.assert_called_once()
@@ -377,7 +377,7 @@ def test_segmentation_dino_success(mocker, mock_sfapi_client, mock_config832):
     assert result is True
 
 
-def test_segmentation_dino_submission_failure(mocker, mock_sfapi_client, mock_config832):
+def test_segmentation_dinov3_submission_failure(mocker, mock_sfapi_client, mock_config832):
     from orchestration.flows.bl832.nersc import NERSCTomographyHPCController
 
     mocker.patch("orchestration.flows.bl832.nersc.time.sleep")
@@ -385,12 +385,12 @@ def test_segmentation_dino_submission_failure(mocker, mock_sfapi_client, mock_co
     mock_sfapi_client.compute.return_value.submit_job.side_effect = Exception("No GPU nodes")
     controller = NERSCTomographyHPCController(client=mock_sfapi_client, config=mock_config832)
 
-    result = controller.segmentation_dino(recon_folder_path="folder/recfile")
+    result = controller.segmentation_dinov3(recon_folder_path="folder/recfile")
 
     assert result is False
 
 
-def test_segmentation_dino_output_paths(mocker, mock_sfapi_client, mock_config832):
+def test_segmentation_dinov3_output_paths(mocker, mock_sfapi_client, mock_config832):
     """
     Output dir should swap /rec for /seg in the folder name and route to /dino.
 
@@ -413,7 +413,7 @@ def test_segmentation_dino_output_paths(mocker, mock_sfapi_client, mock_config83
 
     mock_sfapi_client.compute.return_value.submit_job.side_effect = capture
     controller = NERSCTomographyHPCController(client=mock_sfapi_client, config=mock_config832)
-    controller.segmentation_dino(recon_folder_path="folder/recfile")
+    controller.segmentation_dinov3(recon_folder_path="folder/recfile")
 
     script = captured_scripts[0]
     assert "segfile" in script
@@ -527,32 +527,32 @@ def test_nersc_segmentation_sam3_task_failure(mocker, mock_config832):
     assert result["success"] is False
 
 
-def test_nersc_segmentation_dino_task_success(mocker, mock_config832):
-    from orchestration.flows.bl832.nersc import nersc_segmentation_dino_task
+def test_nersc_segmentation_dinov3_task_success(mocker, mock_config832):
+    from orchestration.flows.bl832.nersc import nersc_segmentation_dinov3_task
 
     mocker.patch("orchestration.flows.bl832.nersc.get_run_logger", return_value=mocker.MagicMock())
     mock_controller = mocker.MagicMock()
-    mock_controller.segmentation_dino.return_value = True
+    mock_controller.segmentation_dinov3.return_value = True
     mocker.patch("orchestration.flows.bl832.nersc.get_controller", return_value=mock_controller)
 
-    result = nersc_segmentation_dino_task.fn(
+    result = nersc_segmentation_dinov3_task.fn(
         recon_folder_path="folder/recfile",
         config=mock_config832
     )
 
-    mock_controller.segmentation_dino.assert_called_once_with(recon_folder_path="folder/recfile")
+    mock_controller.segmentation_dinov3.assert_called_once_with(recon_folder_path="folder/recfile")
     assert result is True
 
 
-def test_nersc_segmentation_dino_task_failure(mocker, mock_config832):
-    from orchestration.flows.bl832.nersc import nersc_segmentation_dino_task
+def test_nersc_segmentation_dinov3_task_failure(mocker, mock_config832):
+    from orchestration.flows.bl832.nersc import nersc_segmentation_dinov3_task
 
     mocker.patch("orchestration.flows.bl832.nersc.get_run_logger", return_value=mocker.MagicMock())
     mock_controller = mocker.MagicMock()
-    mock_controller.segmentation_dino.return_value = False
+    mock_controller.segmentation_dinov3.return_value = False
     mocker.patch("orchestration.flows.bl832.nersc.get_controller", return_value=mock_controller)
 
-    result = nersc_segmentation_dino_task.fn(
+    result = nersc_segmentation_dinov3_task.fn(
         recon_folder_path="folder/recfile",
         config=mock_config832
     )
@@ -594,7 +594,7 @@ def test_nersc_combine_segmentations_task_failure(mocker, mock_config832):
 
 
 # ──────────────────────────────────────────────────────────────────────────────
-# nersc_petiole_segment_flow  (recon + SAM3 + DINO + combine)
+# nersc_petiole_segment_flow  (recon + SAM3 + DINOv3 + combine)
 #
 # Replaces the former nersc_forge_recon_multisegment_flow tests.
 # The cleaned nersc.py exposes nersc_petiole_segment_flow as the canonical
@@ -607,7 +607,7 @@ def test_nersc_combine_segmentations_task_failure(mocker, mock_config832):
 # ──────────────────────────────────────────────────────────────────────────────
 
 def test_petiole_segment_flow_both_succeed(mocker, mock_config832, mock_recon_success):
-    """Recon + SAM3 + DINO all succeed → combine is called → flow returns True."""
+    """Recon + SAM3 + DINOv3 all succeed → combine is called → flow returns True."""
     from orchestration.flows.bl832.nersc import nersc_petiole_segment_flow
 
     mock_controller = mocker.MagicMock()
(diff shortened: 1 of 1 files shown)


davramov:
When reviewing, you can ignore the files in `/scripts/perlmutter/`, I added these for completeness

orchestration/flows/bl832/config.py
```diff
@@ -28,3 +28,5 @@ def _beam_specific_config(self) -> None:
         self.alcf832_scratch = self.endpoints["alcf832_scratch"]
         self.scicat = self.config["scicat"]
         self.ghcr_images832 = self.config["ghcr_images832"]
+        self.nersc_recon_num_nodes = 16
+        self.nersc_segment_num_nodes = 42
```
xiaoyachong: Can we read `nersc_recon_num_nodes` and `nersc_segment_num_nodes` from a config file instead of hardcoding them?

orchestration/flows/bl832/dispatcher.py
```diff
@@ -51,27 +55,37 @@ class DecisionFlowInputModel(BaseModel):
...
     is_export_control: Optional[bool] = Field(default=False)
+    num_nodes: Optional[int] = Field(default=4)
     config: Optional[Union[dict, Any]] = Field(default_factory=dict)
 
 
 @task(name="setup_decision_settings")
-def setup_decision_settings(alcf_recon: bool, nersc_recon: bool, new_file_832: bool) -> dict:
+def setup_decision_settings(
+    alcf_recon: bool,
+    nersc_recon: bool,
```
xiaoyachong, line 66: I was wondering if you could clarify the reason for having both `nersc_recon` and `nersc_recon_multinode`. They seem somewhat similar—are they meant to serve different use cases?

orchestration/flows/bl832/nersc.py
```diff
@@ -123,7 +127,7 @@ def reconstruct(
 #SBATCH --error={pscratch_path}/tomo_recon_logs/%x_%j.err
 #SBATCH -N 1
 #SBATCH --ntasks-per-node 1
-#SBATCH --cpus-per-task 64
+#SBATCH --cpus-per-task 128
```
xiaoyachong: Can we read the values (e.g. `N`, `ntasks-per-node`, `cpus-per-task`) in the `sbatch` command from a config file instead of hardcoding them?

orchestration/flows/bl832/nersc.py
```diff
@@ -196,11 +204,298 @@ def reconstruct(
...
+
+        file_name = f"{path.stem}.h5"
+
+        logger.info(f"File name: {file_name}")
+        logger.info(f"Folder name: {folder_name}")
+        logger.info(f"Number of nodes: {num_nodes}")
+
+        if num_nodes == 8:
+            qos = "debug"
+        if num_nodes < 8:
```
xiaoyachong: may change to `elif`

orchestration/flows/bl832/nersc.py
```diff
@@ -196,11 +204,298 @@ def reconstruct(
...
+            qos = "premium"
+
+# If using with a reservation:
+# SBATCH -q regular
+# SBATCH --reservation=_CAP_MarchModCon_CPU
+# SBATCH -A amsc006
+        # IMPORTANT: job script must be deindented to the leftmost column or it will fail immediately
+        job_script = f"""#!/bin/bash
+#SBATCH -q {qos}
+#SBATCH -A als
```
xiaoyachong: Can we read this value from a config file, since the project (e.g., `als`) may change?

orchestration/flows/bl832/nersc.py
```diff
@@ -238,7 +533,7 @@ def build_multi_resolution(
 #SBATCH --error={pscratch_path}/tomo_recon_logs/%x_%j.err
 #SBATCH -N 1
 #SBATCH --ntasks-per-node 1
-#SBATCH --cpus-per-task 64
+#SBATCH --cpus-per-task 128
```
xiaoyachong: Same as above, could we read this value from a config file so we don’t need to modify the code?

orchestration/flows/bl832/nersc.py
```diff
@@ -292,109 +587,968 @@ def build_multi_resolution(
...
+        logger.info("Starting NERSC segmentation process (inference_v6).")
 
+        user = self.client.user()
+        pscratch_path = f"/pscratch/sd/{user.name[0]}/{user.name}"
+        cfs_path = "/global/cfs/cdirs/als/data_mover/8.3.2"
+        conda_env_path = f"{cfs_path}/envs/sam3-py311"
+
+        # Paths
+        # seg_scripts_dir = f"{cfs_path}/tomography_segmentation_scripts/inference_v4/forge_feb_seg_model_demo/"
+        seg_scripts_dir = f"{cfs_path}/tomography_segmentation_scripts/inference_latest/forge_feb_seg_model_demo/"
```
xiaoyachong: Would it be possible to read `seg_scripts_dir` and `checkpoints_dir` from a config file, as they may differ between projects?

orchestration/flows/bl832/nersc.py
```diff
@@ -292,109 +587,968 @@ def build_multi_resolution(
...
+        checkpoints_dir = f"{cfs_path}/tomography_segmentation_scripts/sam3_finetune/sam3/"
+
+        bpe_path = f"{checkpoints_dir}/bpe_simple_vocab_16e6.txt.gz"
```
xiaoyachong: Not directly related to this PR, but as we collaborate with Harry on a different project, I’m wondering whether there is a more general way to handle model loading across different models.

`MLflow` might be a good option to consider in a future PR (though the inference code would likely need to be adapted accordingly). However, I’m not yet sure how to run a `Slurm` job while loading models from `MLflow`, and there may be some additional configuration required.

orchestration/flows/bl832/nersc.py
```diff
@@ -292,109 +587,968 @@ def build_multi_resolution(
...
+export FT_CKPT="{finetuned_checkpoint}"
+
+srun --ntasks-per-node=1 --gpus-per-task=4 \
+  torchrun \
+    --nnodes={num_nodes} \
+    --nproc_per_node=4 \
+    --rdzv_id=$SLURM_JOB_ID \
+    --rdzv_backend=c10d \
+    --rdzv_endpoint=$MASTER_ADDR:$MASTER_PORT \
+    src/inference_v6.py \
```
xiaoyachong: Could `src/inference_v6.py` and  `prompts` be read from a config file? Also, the `torchrun` command may vary across models. Could we define a function to generate the command for different tasks?

orchestration/flows/bl832/nersc.py
```diff
@@ -292,109 +587,968 @@ def build_multi_resolution(
...
+
+        input_dir = f"{pscratch_path}/8.3.2/scratch/{recon_folder_path}"
+        output_folder = recon_folder_path.replace('/rec', '/seg')
+        output_dir = f"{pscratch_path}/8.3.2/scratch/{output_folder}/sam3"
+
+        logger.info(f"Input directory: {input_dir}")
+        logger.info(f"Output directory: {output_dir}")
+        logger.info(f"Conda environment: {conda_env_path}")
+
+        # Default values (used when defaults=True or variable not found)
```
xiaoyachong: Could we encapsulate lines 622–664 into a function, for example `set_args_values()`?

orchestration/flows/bl832/nersc.py
```diff
@@ -292,109 +587,968 @@ def build_multi_resolution(
...
-        location = "nersc832_alsdev_pscratch_scratch"
+            return {
+                "success": False,
+                "job_id": None,
+                "timing": None,
+                "output_dir": None
+            }
 
-        flow_name = f"delete {location}: {Path(tiff_file_path).name}"
+    def segmentation_dino(
```
xiaoyachong: Similar to `segmentation_sam3()`, we may need to make model loading and inference more flexible to support different projects.

orchestration/flows/bl832/nersc.py
```diff
@@ -292,109 +587,968 @@ def build_multi_resolution(
...
+                    job.complete()
+                    logger.info("DINO segmentation job completed successfully after recovery.")
+                    return True
+                except Exception as recovery_err:
+                    logger.error(f"Failed to recover job {jobid}: {recovery_err}")
+                    return False
+            else:
+                return False
+
+    def combine_segmentations(
```
xiaoyachong, line 964: Same as the other two segmentation functions, we could make this more flexible—for example, by reading `seg_scripts_dir` from a config file.

Also, I wonder if it would be beneficial to move the segmentation and combination code into a separate `segmentation.py` file, as `nersc.py` is getting quite long. In `segmentation.py`, we could load different models depending on the project—for example, for Synaps, we load SAM3 and DINO3 and then combine them, while for Harry’s project, we might load a U-Net model instead.

orchestration/flows/bl832/nersc.py
```diff
@@ -514,9 +1798,501 @@ def nersc_recon_flow(
...
+        logger.info("NERSC reconstruction + segmentation flow completed successfully.")
+        return True
+    else:
+        logger.warning(f"Flow completed with issues: recon={nersc_reconstruction_success}, seg={nersc_segmentation_success}")
+        return False
+
+
+@flow(name="nersc_forge_recon_multisegment_flow",
+      flow_run_name="nersc_recon_multiseg-{file_path}")
+def nersc_forge_recon_multisegment_flow(
```
xiaoyachong: May I know why we define both `nersc_forge_recon_segment_flow` and `nersc_forge_recon_multisegment_flow`?

orchestration/flows/bl832/nersc.py
```diff
@@ -514,9 +1798,501 @@ def nersc_recon_flow(
...
+            file_path=scratch_path_tiff,
+            source=config.nersc832_alsdev_pscratch_scratch,
+            destination=config.data832_scratch
+        )
+        logger.info(f"Transfer reconstructed TIFF data to data832 success: {data832_tiff_transfer_success}")
+    except Exception as e:
+        logger.error(f"Failed to transfer TIFFs to data832: {e}")
+        data832_tiff_transfer_success = False
+
+    # ── STEP 3: SAM3 / DINOv3 ──────────────────────────
```
xiaoyachong, line 1686: This step (SAM3 / DINOv3) is currently hardcoded, but the workflow may vary across different projects. For example, other projects may load/save U-Net features. Could we make this more flexible (perhaps in a future PR)?

scripts/perlmutter/sfapi_reconstruction.py
```diff
@@ -0,0 +1,1755 @@
+from __future__ import print_function
```
xiaoyachong, line 1: I wonder if we need to explain some reconstruction details in our draft manuscript.

orchestration/flows/bl832/dispatcher.py
davramov: The `nersc_recon_flow` is the original flow running in production, which only supported running on 1 CPU node and did not scale. I created `nersc_recon_multinode_flow` as the scalable alternative. For the demo, I did not want to interfere with the stable production settings. Now that we know it's stable, I can fully replace the single-node version from this PR with the multinode implementation.

orchestration/flows/bl832/nersc.py
davramov: `nersc_forge_recon_segment_flow` was the original flow when we were just using sam3, and `nersc_forge_recon_multisegment_flow` combined sam3 and dino. I can remove `nersc_forge_recon_segment_flow`, and rename `nersc_forge_recon_multisegment_flow` to maybe `nersc_petiole_segment_flow`, and start tracking the segmentation flows based on the scientific use case.

orchestration/flows/bl832/nersc.py
davramov: I agree, this script (and alcf.py) will become quite large fairly quickly as we add more methods. Not for this PR, but the next one where we integrate Harry's code, I think we can split it up so ALCF/NERSC flows live in their own subdirectories:

```
flows/bl832/
    nersc/
        __init__.py          # re-exports NERSCTomographyHPCController
        controller.py        # class definition + reconstruct, build_multi_resolution
        segmentation.py      # segmentation_sam3, segmentation_dino, combine_segmentations
        streaming.py         # streaming mixin stuff
    alcf/
        __init__.py
        controller.py
        segmentation.py
```

orchestration/flows/bl832/nersc.py
davramov: I agree, we should address this in another PR. I'll create a new issue that captures this and other larger changes that seem out of scope for getting this PR merged

orchestration/flows/bl832/nersc.py
davramov: I'll put this suggestion in a new issue with the others.

orchestration/flows/bl832/nersc.py
davramov: I included the first suggestion, but for the `torchrun` command part, I'll put it in a new issue with the others.

orchestration/flows/bl832/config.py
```diff
@@ -26,5 +27,12 @@ def _beam_specific_config(self) -> None:
...
         self.alcf832_raw = self.endpoints["alcf832_raw"]
         self.alcf832_scratch = self.endpoints["alcf832_scratch"]
+        # SciCat
         self.scicat = self.config["scicat"]
+        # NERSC HPC submission settings
         self.ghcr_images832 = self.config["ghcr_images832"]
+        self.nersc_recon_settings = self.config["hpc_submission_settings832"]["nersc_reconstruction"]
+        self.nersc_multiresolution_settings = self.config["hpc_submission_settings832"]["nersc_multiresolution"]
+        self.nersc_segment_sam3_settings = self.config["hpc_submission_settings832"]["nersc_segmentation_sam3"]
+        self.nersc_segment_dino_settings = self.config["hpc_submission_settings832"]["nersc_segmentation_dino"]
```
xiaoyachong: It’s not critical, but the correct version is DINOv3. That said, using the names “sam” and “dino” should still work, since the inference routine is consistent across different versions.

orchestration/flows/bl832/dispatcher.py
xiaoyachong: Thanks for your information!

orchestration/flows/bl832/nersc.py
xiaoyachong: Sounds great

orchestration/flows/bl832/nersc.py
```diff
@@ -546,14 +1867,181 @@ def nersc_streaming_flow(
...
+        logger.info(f"Segmentation success: {success}")
+    else:
+        success = bool(nersc_segmentation_success)
+    if not success:
+        logger.error("Segmentation Failed.")
+    return nersc_segmentation_success
+
+
+@task(name="nersc_segmentation_dino_task")
+def nersc_segmentation_dino_task(
```
xiaoyachong: I noticed that on NERSC there are `segmentation_dino` and `nersc_segmentation_dino_task`, while on ALCF there are `_segmentation_dino_wrapper`, `segmentation_dino`, and `alcf_segmentation_dino_task`. Could we make these naming conventions consistent across both environments?

scripts/perlmutter/sfapi_reconstruction.py
xiaoyachong: It seems Prady will add those details in his paper.

orchestration/flows/bl832/nersc.py
davramov: The `_segmentation_dino_wrapper` is necessary to submit to Globus Compute, but not needed for SFAPI. Otherwise I'll make sure it is consistent when I continue working on the ALCF implementation.

PR #132, opened 2026-04-06
Issue 131: MLflow integration

davramov:
# New files

`orchestration/mlflow.py` — Core MLflow integration module with:
- `ModelCheckpointInfo` dataclass holding checkpoint paths and inference params
- `_is_mlflow_reachable()` — fast health check (2s timeout) to short-circuit before MLflow's built-in 30s retry loop fires on a down server

- `get_mlflow_client()` — constructs a client from `config.mlflow["tracking_uri"]`
- `get_checkpoint_info()` — resolves a model's production alias from the registry; returns None gracefully on any failure, allowing callers to fall back to config defaults
- `register_checkpoint()` — registers a model version with `nersc_path`, `alcf_path,` and inference params as version-level tags; list/dict values are JSON-encoded automatically
- `log_segmentation_metrics()` — logs timing and job params as an (optionally nested) MLflow run after SLURM job completion on the orchestration server, not the compute node

`orchestration/flows/bl832/register_mlflow.py`
- One-time registration script for `sam3-petiole` and `dinov3-petiole,` storing all model-coupled params (checkpoint paths, conda env, scripts dir, batch size, prompts, confidence, overlap) in the registry.
- Includes  `retrieve_mlflow_params_test()` to validate the resolution chain end-to-end, with separate checks confirming MLflow-coupled params are overridden and SLURM allocation params are unchanged.

# Modified files

`orchestration/flows/bl832/config.py`
- Added `self.mlflow = self.config["mlflow"]["local"]` to `Config832._beam_specific_config(),` exposing the MLflow environment dict (containing `tracking_uri` and `registry_uri`)
- `config.yml` — Added a multi-environment mlflow section with four named environments:
```yaml
# `Config832` defaults to local; switching environments requires only changing the key in `_beam_specific_config`.  
mlflow:
    local:     # http://localhost:5001
    dev:       # http://mlflow-dev.computing.als.lbl.gov
    prod:      # https://mlflow.computing.als.lbl.gov
    staging:   # https://mlflow-staging.computing.als.lbl.gov
```
  
`orchestration/flows/bl832/nersc.py` — Refactored `_load_job_options()` to implement a three-layer resolution:

  1. Config YAML (base defaults — SLURM allocation, infrastructure)
  2. MLflow Registry (model-coupled params — checkpoints, inference hyperparams, envs, scripts) — skipped silently if mlflow_model_name is not passed or server is unreachable
  3. Prefect Variable (hot overrides — wins over everything, no redeploy needed)

`segmentation_sam3()` and `segmentation_dinov3()` call `_load_job_options` with `mlflow_model_name` and `mlflow_checkpoint_key` to activate the MLflow layer.

### DIFF
--- a/orchestration/flows/bl832/register_mlflow.py
+++ b/orchestration/flows/bl832/register_mlflow.py
@@ -1,0 +1,195 @@
+import logging
+
+from orchestration.flows.bl832.config import Config832
+from orchestration.flows.bl832.nersc import _load_job_options
+from orchestration.mlflow import register_checkpoint
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def register_mlflow_checkpoints():
+    config = Config832()
+
+    scripts_dir = "/global/cfs/cdirs/als/data_mover/8.3.2/tomography_segmentation_scripts/"
+
+    register_checkpoint(
+        model_name="sam3-petiole",
+        nersc_path=f"{scripts_dir}sam3_finetune/sam3/checkpoint_v6.pt",
+        alcf_path="/eagle/IRIBeta/als/seg_models/sam3/checkpoint_v6.pt",
+        config=config,
+        alias="production",
+        description="SAM3 v6 fine-tuned on petiole micro-CT data.",
+        inference_params={
+            # ── paths ──────────────────────────────────────────────────────────
+            "original_checkpoint_path":
+            f"{scripts_dir}sam3_finetune/sam3/sam3.pt",
+            "bpe_path": f"{scripts_dir}sam3_finetune/sam3/bpe_simple_vocab_16e6.txt.gz",
+            "conda_env_path": "/global/cfs/cdirs/als/data_mover/8.3.2/envs/sam3-py311",
+            "seg_scripts_dir": f"{scripts_dir}inference_latest/forge_feb_seg_model_demo/",
+            "checkpoints_dir": f"{scripts_dir}sam3_finetune/sam3/",
+            # ── inference hyperparameters ───────────────────────────────────────
+            "script_name": "src/inference_v6.py",
+            "batch_size": 1,
+            "patch_size": 400,
+            "confidence": [0.5],       # list → JSON-encoded automatically
+            "overlap": 0.25,
+            "prompts": [               # list → JSON-encoded automatically
+                "Phloem Fibers",
+                "Hydrated Xylem vessels",
+                "Air-based Pith cells",
+                "Dehydrated Xylem vessels",
+            ],
+        },
+    )
+
+    register_checkpoint(
+        model_name="dinov3-petiole",
+        nersc_path="/global/cfs/cdirs/als/data_mover/8.3.2/tomography_segmentation_scripts/dino/best.ckpt",
+        alcf_path="/eagle/IRIBeta/als/seg_models/dino/best.ckpt",
+        config=config,
+        alias="production",
+        description="DINOv3 fine-tuned on petiole micro-CT data.",
+        inference_params={
+            # ── paths ──────────────────────────────────────────────────────────
+            "conda_env_path": "/global/cfs/cdirs/als/data_mover/8.3.2/envs/dino_demo",
+            "seg_scripts_dir": f"{scripts_dir}inference_v5_multiseg/forge_feb_seg_model_demo/",
+            # ── inference hyperparameters ───────────────────────────────────────
+            "script_name": "src.inference_dino_v1",
+            "batch_size": 4,
+            "nproc_per_node": 4,
+        },
+    )
+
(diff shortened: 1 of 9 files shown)


orchestration/flows/bl832/register_mlflow.py
```diff
@@ -0,0 +1,195 @@
...
+
+from orchestration.flows.bl832.config import Config832
+from orchestration.flows.bl832.nersc import _load_job_options
+from orchestration.mlflow import register_checkpoint
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def register_mlflow_checkpoints():
```
xiaoyachong, line 11: The model registration code is clean and concise. I think it is good for now. However, in the future, we may want to define a unified format for all registered models. Meanwhile, we are also planning to build a web-based UI for model registration. We intend to hire an intern for this.

FYI, for latent space exploration, the strategy we adopt is to create an MLflow model wrapper and implement a `predict()` API. The main benefit is that different ML models can expose different inference interfaces. For example, to obtain latent features from a ViT, we may use `latent, _ = self.model.encoder(tensor)`, while for a VAE we may use `recon_batch, mu, logvar = model(inputs)`. In addition, non-ML models such as UMAP and PCA cannot be used directly in the same way. As a result, switching between models may otherwise require changes in the application code.

To avoid that, it would be beneficial to have a unified inference API, and an [MLflow model wrapper](https://mlflow.org/docs/latest/api_reference/python_api/mlflow.pyfunc.html) is a good way to achieve this.


Here is some LSE code for reference:
* ML model registration with an MLflow wrapper: `https://github.com/mlexchange/mlex_latent_explorer/tree/main/live_operator_example`
* Application code for inference: `https://github.com/mlexchange/mlex_latent_explorer/blob/main/src/arroyo_reduction/reducer.py`
* `mlflow_utils` for loading a model: `https://github.com/mlexchange/mlex_utils/blob/main/mlex_utils/mlflow_utils/mlflow_model_client.py`

orchestration/flows/bl832/register_mlflow.py
```diff
@@ -0,0 +1,195 @@
...
+            "seg_scripts_dir": f"{scripts_dir}inference_v5_multiseg/forge_feb_seg_model_demo/",
+            # ── inference hyperparameters ───────────────────────────────────────
+            "script_name": "src.inference_dino_v1",
+            "batch_size": 4,
+            "nproc_per_node": 4,
+        },
+    )
+
+
+def retrieve_mlflow_params_test() -> bool:
```
xiaoyachong, line 65: I wonder if you want to re-use the mlflow_utils code: https://github.com/mlexchange/mlex_utils/blob/main/mlex_utils/mlflow_utils/mlflow_model_client.py

orchestration/mlflow.py
```diff
@@ -0,0 +1,281 @@
...
+import mlflow
+from mlflow.tracking import MlflowClient
+
+from orchestration.config import BeamlineConfig
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ModelCheckpointInfo:
```
xiaoyachong, line 16: may re-use mlflow_utils: https://github.com/mlexchange/mlex_utils/blob/main/mlex_utils/mlflow_utils/mlflow_model_client.py

orchestration/mlflow.py
```diff
@@ -0,0 +1,281 @@
...
+        model_name=model_name,
+        version=mv.version,
+        alias=alias,
+        nersc_path=nersc_path,
+        alcf_path=alcf_path,
+        inference_params=inference_params,
+    )
+
+
+def register_checkpoint(
```
xiaoyachong, line 146: As mentioned earlier, we can register models using an [MLflow wrapper](https://mlflow.org/docs/latest/api_reference/python_api/mlflow.pyfunc.html) in the future. 
```
class MyModel(mlflow.pyfunc.PythonModel):
    def predict(self, context, model_input):
        return ...
```
Meanwhile, the predict() method in an MLflow model wrapper (pyfunc) is designed to serve as a unified inference interface. 
`mlflow models serve -m runs:/<run_id>/model -p 5000`
With this in place, we can easily extend it to [MLflow inference endpoints](https://mlflow.org/docs/3.11.1/ml/deployment/deploy-model-locally/).

xiaoyachong (commented):
Thanks for your PR—it looks good and is ready to merge. We can use it for Harry’s beamtime.

I’ve left a few comments with suggestions for future refactoring (e.g., [MLflow model wrapper](https://mlflow.org/docs/latest/api_reference/python_api/mlflow.pyfunc.html), predict() API, and [inference endpoints](https://mlflow.org/docs/3.11.1/ml/deployment/deploy-model-locally/)).

PR #133, opened 2026-04-06
[8.3.2] Moon rock segmentation

davramov:
This PR incorporates the models and inference code from https://github.com/als-computing/forge_feb_seg_model_demo/pull/7 to support Harry's moon rock segmentation.
- generalizes DINOv3 code to accept different `project` names ("petiole" or "moon")
- creates a new `nersc_moon_segment_flow()`

To do:
- add a block to `register_mlflow_checkpoints()` for moon settings
- add flow to `dispatcher.py` settings

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -175,6 +175,7 @@ mlflow:
     registry_uri: https://mlflow-staging.computing.als.lbl.gov
 
 hpc_submission_settings832:
+  # ── RECON + MULTIRES SETTINGS ───────────────────────────────────────────────
   nersc_reconstruction:
     # ── SLURM resource allocation ─────────────────────────────────────────────
     qos: realtime
@@ -190,6 +191,8 @@ hpc_submission_settings832:
     reservation: ""
     cpus-per-task: 128
     walltime: "0:15:00"
+
+  # ── PETIOLE SEGMENTATION SETTINGS ───────────────────────────────────────────
   nersc_segmentation_sam3:
     # ── SLURM resource allocation ─────────────────────────────────────────────
     qos: regular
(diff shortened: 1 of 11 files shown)


orchestration/flows/bl832/config.py
```diff
@@ -29,10 +29,13 @@ def _beam_specific_config(self) -> None:
         self.alcf832_scratch = self.endpoints["alcf832_scratch"]
         # SciCat
         self.scicat = self.config["scicat"]
+        # MLflow
+        self.mlflow = self.config["mlflow"]["local"]
         # NERSC HPC submission settings
         self.ghcr_images832 = self.config["ghcr_images832"]
         self.nersc_recon_settings = self.config["hpc_submission_settings832"]["nersc_reconstruction"]
         self.nersc_multiresolution_settings = self.config["hpc_submission_settings832"]["nersc_multiresolution"]
         self.nersc_segment_sam3_settings = self.config["hpc_submission_settings832"]["nersc_segmentation_sam3"]
```
xiaoyachong, line 38: I was wondering if we should add petiole or synaps-i to `nersc_segment_sam3_settings`, `nersc_segment_dinov3_settings`, and `nersc_combine_segmentation_settings`.

orchestration/flows/bl832/register_mlflow.py
```diff
@@ -0,0 +1,251 @@
...
+
+from orchestration.flows.bl832.config import Config832
+from orchestration.flows.bl832.nersc import _load_job_options
+from orchestration.mlflow import register_checkpoint
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def register_mlflow_checkpoints():
```
xiaoyachong, line 11: I’m wondering if this is separate from the main workflow—do we register the models before any flow is triggered?

Also, do we need to set up a dedicated conda environment for model registration, or can we simply add MLflow to the existing `requirements.txt` or `pyproject.toml`? We may also need to include related instructions in the `README`.

config.yml
```diff
@@ -160,7 +160,22 @@ prefect:
...
 
+mlflow:
+  local:
+    tracking_uri: http://localhost:5001
+    registry_uri: http://localhost:5001
+  dev:
+    tracking_uri: http://mlflow-dev.computing.als.lbl.gov
+    registry_uri: http://mlflow-dev.computing.als.lbl.gov
+  prod:
+    tracking_uri: https://mlflow.computing.als.lbl.gov
```
xiaoyachong, line 171: I’m wondering how you register models to the production/staging MLflow server instead of the local one. I’ve been using:
```
from als_computing import enable_auth
enable_auth()
```

However, the [MLFLOW_TRACKING_TOKEN](https://mlflow.org/docs/3.11.1/self-hosting/architecture/tracking-server/#authentication-and-encryption) method didn’t work during my last test with Dylan. I’m wondering if there have been any updates.

Also, I’m wondering if you would need to update the `.env` file to include the authentication details.

pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "mkdocs",
     "mkdocs-material",
     "mkdocs-mermaid2-plugin",
+    "mlflow==2.22.0",
```
xiaoyachong, line 17: It seems the MLflow registration environment is the same as the existing one. I’m wondering whether we need a separate environment or not.

xiaoyachong (commented):
Thanks for your PR. It looks great and I leave a few comments.

config.yml
davramov: I'll work with Dylan to get connected to the production server. He mentioned there is a token he can share with me that will work with authentication

pyproject.toml
davramov: Hmm, I can split MLflow out into an optional dependency

orchestration/flows/bl832/register_mlflow.py
davramov: Registering the models before triggering the flows was my thought.

But, I wonder if it makes sense to dynamically re-register models to MLflow. Say, for example, we update config.yml or the Prefect variables with different model parameters, maybe we want to track those changes for future reference.

PR #134, opened 2026-04-21
[832] Copy to beegfs and register in Tiled

davramov:
This PR adds a transfer for the bl832 workflow to copy data to beegfs, and ingest the new data into Tiled.


orchestration/tiled.py
```diff
@@ -0,0 +1,76 @@
...
+        path: Absolute path on the client filesystem (used for logging).
+        prefix: Optional sub-path within the Tiled catalog.
+        overwrite: Whether to overwrite existing entries in the Tiled catalog.
+    """
+    logger = get_run_logger()
+    load_dotenv()
+    tiled_uri = os.environ["TILED_URI"]
+    api_key = os.environ["TILED_SINGLE_USER_API_KEY"]
+
+    client = from_uri(tiled_uri, api_key=api_key)
```
dylanmcreynolds: You should omit the api_key from this call. Tiled picks up TILED_API_KEY from the environment by default. You have this code in there because you're using the easiest method to start tiled on a dev server, which takes the TILED_SINGLE_USER_API_KEY. That variable is not used in production systems.

orchestration/tiled.py
```diff
@@ -0,0 +1,76 @@
...
+    logger = get_run_logger()
+    path = Path(path)
+    logger.info(f"Submitting task: register {path} to Tiled (prefix={prefix!r})")
+    await register_file_to_tiled(path, prefix=prefix, overwrite=overwrite)
+
+
+if __name__ == "__main__":
+    import asyncio
+
+    zarr = Path("/Users/david/Documents/data/tomo/scratch/rec20230606_152011_jong-seto_fungal-mycelia.zarr")
```
dylanmcreynolds: Maybe don't leave your system path in the code. Use env vars?

requirements.txt
```diff
@@ -17,3 +17,5 @@ python-dotenv
 pyyaml
 scicat-beamline @ git+https://github.com/als-computing/scicat_beamline.git@4828273f5f49ba4eba5442728729e0545b3f5b79
 sfapi_client
+tiled[client]
```
dylanmcreynolds: We should remvoe requirements.txt from this project...as it's a duplicate of what is now in pyproject.toml.

pyproject.toml
```diff
@@ -22,7 +22,9 @@ dependencies = [
     "python-dotenv",
     "pyyaml",
     "scicat_beamline @ git+https://github.com/als-computing/scicat_beamline.git@4828273f5f49ba4eba5442728729e0545b3f5b79",
```
dylanmcreynolds, line 26: This this PR in scicat_beamline mergable?

dylanmcreynolds (changes requested):

orchestration/tiled.py
```diff
@@ -0,0 +1,180 @@
...
+    if not tags:
+        return
+
+    def _apply_tags(entry_node):
+        existing_blob = entry_node.access_blob
+        existing_tags = (existing_blob or {}).get("tags", [])
+        merged_tags = list(set(existing_tags) | set(tags))
+        op = "replace" if existing_blob is not None else "add"
+        try:
+            entry_node.patch_metadata(
```
dylanmcreynolds: I think this could be a one-liner, without building the whole patch object:
```
entry.update_metadata(access_tags=merged_tags)
```

orchestration/tiled.py
```diff
@@ -0,0 +1,180 @@
...
+                access_blob_patch=[{"op": op, "path": "", "value": {"tags": merged_tags}}],
+            )
+            logger.info(f"Tagged {entry_node.uri} with {merged_tags}")
```
dylanmcreynolds: I would be really careful with logging in a recursive function...it could go nuts.  I'd change these all to logger.debug. 

you can also save some processing by only producing the log string if debug is enabled:

```
if logger.isEnabledFor(logging.DEBUG):
   logger.debug(f"don't build a new string on every pass when doing an f-string {value}")
```

orchestration/tiled.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Returns:
+        A tuple ``(ok, actual_tags)`` where ``ok`` is True iff every tag in
+        ``expected_tags`` is present in the entry's ``access_blob.tags``.
+
+    Raises:
+        KeyError: If the entry cannot be located under ``prefix``.
+    """
+    logger = get_run_logger()
+    load_dotenv()
```
dylanmcreynolds: load_dotenv() should be outside functions...run when the module loads

orchestration/tiled.py
```diff
@@ -0,0 +1,180 @@
...
+        # TIFF directory: Tiled registers each file flat into the prefix node
+        for key in node:
+            _apply_tags(node[key])
+    else:
+        # .h5 or .zarr: registered as single node, keyed by stem
+        # Even on COLLISION the entry exists — just try it directly
+        entry_key = path.stem
+        logger.info(f"Looking up entry key {entry_key!r} under {prefix!r}")
+        try:
+            _apply_tags(node[entry_key])
```
dylanmcreynolds: if `_apply_tags` is called in both if and else here...what keep this from being an infinite loop? Keyerror? SWhould you  breaking out when you're setting the tag at the top of this function?

pyproject.toml
davramov: I'll have to test it again with the main branch. I remember pinning it while Garrett was adding enhancements to scicat_beamline, and I was nervous about it affecting beamtime.

orchestration/tiled.py
davramov: I'll test this out. It looks cleaner than what I came up with.

davramov: Hmm, I have run through this a few times locally and I'm guessing there must be a difference between how I have tiled set up locally vs what we expect from production.

If I run with:

```
entry_node.patch_metadata(
    access_blob_patch=[{"op": op, "path": "", "value": {"tags": merged_tags}}],
)
```

I get:

```python
14:06:29.096 | INFO    | Task run 'check-tags-/Users/david/Documents/data/tomo/raw/20241216_154449_ddd.h5' - 20241216_154449_ddd.h5 under 'beamlines/bl832/raw/': expected=['bl832'] actual=['bl832'] ok=True
✓ 20241216_154449_ddd.h5: expected=['bl832'] actual=['bl832']
14:06:29.211 | INFO    | Task run 'check-tags-/Users/david/Documents/data/tomo/rec20230224_132553_sea_shell' - rec20230224_132553_sea_shell under 'beamlines/bl832/scratch': expected=['bl832', 'dabramov'] actual=['dabramov', 'bl832'] ok=True
✓ rec20230224_132553_sea_shell: expected=['bl832', 'dabramov'] actual=['dabramov', 'bl832']
14:06:29.296 | INFO    | Task run 'check-tags-/Users/david/Documents/data/tomo/scratch/rec20230606_152011_jong-seto_fungal-mycelia.zarr' - rec20230606_152011_jong-seto_fungal-mycelia.zarr under 'beamlines/bl832/scratch': expected=['bl832'] actual=['bl832'] ok=True
```

And if I run with:

```python
entry_node.update_metadata(access_tags=merged_tags)
```

I get these errors:

```python
14:09:38.889 | INFO    | Task run "apply-tags-['dabramov', 'bl832']" - Could not tag http://127.0.0.1:8000/api/v1/metadata/beamlines/bl832/scratch/20230224_132553_sea_shell_: 403: Access policy rejects the provided access blob.
access_blob must be in the form '{"tags": ["tag1", "tag2", ...]}'
Received access_blob={'user': 'dabramov', 'tags': ['dabramov', 'bl832']}
If this was a merge patch on a user-owned node, use a replace op instead. http://127.0.0.1:8000/api/v1/metadata/beamlines/bl832/scratch/20230224_132553_sea_shell_
```

```python
14:09:40.791 | INFO    | Task run 'check-tags-/Users/david/Documents/data/tomo/raw/20241216_154449_ddd.h5' - 20241216_154449_ddd.h5 under 'beamlines/bl832/raw/': expected=['bl832'] actual=[] ok=False
✗ 20241216_154449_ddd.h5: expected=['bl832'] actual=[]
14:09:40.893 | INFO    | Task run 'check-tags-/Users/david/Documents/data/tomo/rec20230224_132553_sea_shell' - rec20230224_132553_sea_shell under 'beamlines/bl832/scratch': expected=['bl832', 'dabramov'] actual=[] ok=False
✗ rec20230224_132553_sea_shell: expected=['bl832', 'dabramov'] actual=[]
14:09:40.972 | INFO    | Task run 'check-tags-/Users/david/Documents/data/tomo/scratch/rec20230606_152011_jong-seto_fungal-mycelia.zarr' - rec20230606_152011_jong-seto_fungal-mycelia.zarr under 'beamlines/bl832/scratch': expected=['bl832'] actual=[] ok=False
✗ rec20230606_152011_jong-seto_fungal-mycelia.zarr: expected=['bl832'] actual=[]
```

davramov: Here is my tiled config:

```yaml
trees:
- path: /
  tree: catalog
  args:
    uri: /tmp/catalog.db
    writable_storage:
    - /Users/david/Documents/data
    readable_storage:
    - /Users/david/Documents/data
    init_if_not_exists: true
    top_level_access_blob: { "tags": [ "singleuser" ] }

authentication:
  providers:
  - provider: toy
    authenticator: tiled.authenticators:DictionaryAuthenticator
    args:
      users_to_passwords:
        dabramov: ${TILED_PASSWORD}
  tiled_admins:
  - provider: toy
    id: dabramov
  secret_keys:
  - ${TILED_API_KEY}

database:
  uri: sqlite:////tmp/authdb.db
  init_if_not_exists: true

access_control:
  access_policy: "tiled.access_control.access_policies:TagBasedAccessPolicy"
  args:
    provider: toy
    scopes:
    - read:metadata
    - read:data
    - write:metadata
    - write:data
    - create:node
    - register
    - delete:node
    - delete:revision
    tags_db:
      uri: "file:///deploy/tags/tags.db"
    access_tags_parser: "tiled.access_control.access_tags:AccessTagsParser"

uvicorn:
  host: 0.0.0.0
  port: 8000
```

and the docker command for starting up:
```bash
docker stop tiled-server && docker rm tiled-server
rm catalog.db && touch catalog.db
docker run -d \
  -p 127.0.0.1:8000:8000 \
  --env-file .env \
  -v $(pwd)/config.yml:/deploy/config/config_for_containerfile.yml \
  -v $(pwd)/tags.db:/deploy/tags/tags.db \
  -v $(pwd)/catalog.db:/tmp/catalog.db \
  -v /Users/david/Documents/data:/Users/david/Documents/data:ro \
  --name tiled-server \
  ghcr.io/bluesky/tiled:0.2.9
  ```

File: orchestration/tiled.py (hunk 8 above)
Comment (davramov) [reply orchestration/tiled.py]: I'm moving `_apply_tags` to its own task. It only loops if it is a directory (i.e., TIFF stack) to apply the tag to all of the files, otherwise, it only applies the tag to the top level.

```
@task(name="apply-tags", task_run_name="apply-tags-{tags}")
def _apply_tags(entry_node, tags: list[str]) -> None:
```

But this is still where I'm having trouble with 

```
entry_node.update_metadata(access_tags=merged_tags)
```
vs
```
entry_node.patch_metadata(
    access_blob_patch=[{"op": op, "path": "", "value": {"tags": merged_tags}}],
)
```

dylanmcreynolds (approved):

